PR #6, opened 2018-12-07
Enhance cmake envirionment

hjmjohnson:
A review and modernization of the cmake build environment.


hjmjohnson:
NOTE:  Windows Shared libraries (dll's) probably don't work with this code.  Making shared libraries work on windows requires special care for all exported symbols.

I know what needs to be done, but it will make a dependency that only cmake builds would be supported after the change.

afni-rickr:
I just added some final newline characters.

hjmjohnson:
@afni-rickr Thanks.  I squashed those into one commit to keep the history a little cleaner.  I appreciate your review.

PR #13, opened 2018-12-12
WIP:    Change build

leej3:
Build some additional targets including
    nifti_tool, cifti_tool,
    and the nifti2io library.

    Use the properties of targets to handle
    transitive dependencies


CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
-cmake_minimum_required(VERSION 2.8.12 FATAL_ERROR)
-  
+cmake_minimum_required(VERSION 3.11 FATAL_ERROR)
```
hjmjohnson: I'd recommend 3.10.2 as that is what is distributed by default in Ubuntu 18.04 and is rapidly becoming a baseline standard for CI environments.

CMakeLists.txt
```diff
@@ -32,32 +32,58 @@ endif()
...
 
-add_definitions(-DHAVE_ZLIB)
+include(FetchContent)
+include(GNUInstallDirs)
+set(INSTALL_CONFIGDIR ${CMAKE_INSTALL_LIBDIR}/cmake/${NIFTI_NIFTILIB_NAME})
+
 
 # add option to build shared libraries. The default is OFF to maintain the
 # current build behavior
+set( 1)
```
hjmjohnson: What is intended by the set(1) call?

CMakeLists.txt
```diff
@@ -32,32 +32,58 @@ endif()
...
 
-###  USE AS PART OF ITK
+##  USE AS PART OF ITK
 if("${ITK_VERSION_MAJOR}" GREATER 0.0)
 #   message(STATUS "---------------------ITK ZLIB  ${ITK_ZLIB_LIBRARIES}")
   add_definitions(-DITKZLIB)
   set(PACKAGE_PREFIX "ITK")
   set(NIFTI_LIBRARY_PROPERTIES ${ITK_LIBRARY_PROPERTIES})
-  set(NIFTI_INSTALL_BIN_DIR "${ITK_INSTALL_BIN_DIR_CM24}")
-  set(NIFTI_INSTALL_LIB_DIR "${ITK_INSTALL_LIB_DIR_CM24}")
```
hjmjohnson: Why is this removed?

CMakeLists.txt
```diff
@@ -32,32 +32,58 @@ endif()
...
-
-  if(ITK_USE_FLAT_DIRECTORY_INSTALL)
-    set(NIFTI_INSTALL_INCLUDE_DIR "${ITK_INSTALL_INCLUDE_DIR_CM24}")
```
hjmjohnson: Why is this removed?

fsliolib/CMakeLists.txt
```diff
@@ -1,36 +1,43 @@
...
+target_link_libraries( ${NIFTI_LIB_NAME} ${PACKAGE_PREFIX}niftiio)
 if(NOT NIFTI_INSTALL_NO_LIBRARIES)
-  install(TARGETS ${NIFTI_FSLIOLIB_NAME}
+  install(TARGETS ${NIFTI_LIB_NAME}
           EXPORT NIFTITargets
-    RUNTIME DESTINATION ${NIFTI_INSTALL_BIN_DIR} COMPONENT RuntimeLibraries
-    LIBRARY DESTINATION ${NIFTI_INSTALL_LIB_DIR} COMPONENT RuntimeLibraries
-    ARCHIVE DESTINATION ${NIFTI_INSTALL_LIB_DIR} COMPONENT Development)
+    RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR} COMPONENT RuntimeLibraries
+    LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR} COMPONENT RuntimeLibraries
```
hjmjohnson: This will likely break integration with ITK.  I'll need more time to review.

znzlib/CMakeLists.txt
```diff
@@ -1,35 +1,34 @@
 #File Generated by Hans J. Johnson
 #Please contact hans-johnson@uiowa.edu for making enhancments/corrections
-set(ZNZLIB_SRC znzlib.c)
-include_regular_expression("^.*$")
+cmake_minimum_required(VERSION 3.11 FATAL_ERROR)
```
hjmjohnson: minimum required version is only needed at the topmost level.

leej3:
@hjmjohnson I wonder could I get some feedback on this. I am in the process of modifying the AFNI build system and the nifti library seemed like a good test case for attempting to generate a build that was modular and easily enable downstream cmake users to linking against targets from within the sub-project.

This still has some messy bits. I haven't worked out which tests should be added, I have certainly broken some of the functionality. I have also broken the NIFTI...DIR setting and functionality as well as the PACKAGE_PREFIX usage for now. 

Do you think something along these lines would be a useful change for this project though? As far as I can tell the FetchContent functionality of cmake version> 3.11 enables easier management of modules within a subproject. Is this too high a version to constrain oneself to for too little pay off though?

 Using target properties etc. for the linking like I have done here would allow the resolution of transitive dependencies a little more easily which seems nice. I am considering this sort of structure for my rewrite of the AFNI build. I have been looking around for other nice examples of using cmake. ITK certainly is one of them that I see you have experience with. I struggled to intuit all of the inner workings of its approach though, and my aim was to touch the existing AFNI codebase as little as possible...

Any suggestions would be much appreciated...

Cheers

CMakeLists.txt
leej3: Overall my attempt here was to use "include(GnuInstallDirs)"  to more concisely set installation/include/binary directories in each sub-directory so that they could be built individually or could be built as part of the full library. I'm happy to undo this, I just didn't want to verbosely copy with the non-existence of these directories for each sub-build. Perhaps the ability to build each sub-directory is not at all desirable though and so worth this substantial change in existing behaviour (plus I'm not sure how to implement similar behaviour for windows)

CMakeLists.txt
leej3: I have not fully figured out the installation yet. The various binaries and libraries happily install but the linking is broken. I have a separate build working that linked fine before and after installation. I was going to work through what I did differently tomorrow. Once I have that fixed I can happily restored the installation functionality for ITK/non-ITK installations once I have this working

leej3:
Apologies. I didn't send my comment on why the changes quickly enough! Let me know if you have any questions. I have applied the changes a little more gradually (though I fear still not sufficiently clearly on my fork https://github.com/leej3/nifti_clib/commits/explore_use_of_targets_and_fetchcontents).

Overall I just wanted to see if you thought these changes 
a) would be useful to make in some form to the build
b) would be a useful approach to modularize a build in general.

I'm happy to make the changes required, undo some of the things I've broken if you think it is a useful approach worth using here.

Thanks

John

CMakeLists.txt
hjmjohnson: GnuInstallDirs is very unix /linux centric.  Mac bundles and windows don't follow those same conventions.

For tools that I know are only needed on unix, it is OK.  I'd prefer not use GnuInstallDirs for nifti.

CMakeLists.txt
hjmjohnson: This whole section should be protected by the 

if("${ITK_VERSION_MAJOR}" GREATER 0.0)  conditional and should not affect your normal builds.

znzlib/CMakeLists.txt
hjmjohnson: if you do not intend to build zlib separately, it would not be necessary here.

hjmjohnson:
I think that using cmake 3.11 or greater for AFNI and then using the "FetchContent" functionality there is a great idea.  Niftilib, however does not need that feature, and it breaks integration with other packages.

CMakeLists.txt
```diff
@@ -32,32 +32,58 @@ endif()
...
+INSERT_INTO_DIRMAP(utils utils)
+INSERT_INTO_DIRMAP(znz znzlib)
+INSERT_INTO_DIRMAP(nifti2io nifti2)
+INSERT_INTO_DIRMAP(niftiio niftilib)
+INSERT_INTO_DIRMAP(cifti cifti)
+INSERT_INTO_DIRMAP(nifticdf nifticdf)
+
+# Fetch and populate the content of each sub-directory to provide access to
+# targets during cmake configure time
+foreach(module nifticdf niftiio nifti2io cifti)
```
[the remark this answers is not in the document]
leej3: In order to link to the libraries (in this case libraries in the sub-directories) I thought I required the FetchContent commands. From the cmake FetchContent documentation it seems like this file is a good place to define the behavior of the building of the other libraries. As it stands, I think this can be left out, I just couldn't figure out the pattern to implement that.

znzlib/CMakeLists.txt
leej3: znz sometimes uses zlib but AFNI doesn't build it so the cmake find module is fine for linking against it.

Some of the code in the cmake files of the sub directories was to enable individual builds of those sub-directories. I shall abandon that for the sake of being concise

fsliolib/CMakeLists.txt
leej3: Sure that's fine. If the gnuinstalldirs module doesn't have an equivalently convenient analog on mac and windows then its not worth it and abandoning this change for the original NIFTI_INSTALL...DIR variables is better.

hjmjohnson:
This work has been superseded by two other PR's #14  #16

PR #16, opened 2018-12-13
Update cmake 3.10.2

leej3:
These are possible changes to the build. Merge as required/decided.


leej3:
Great. Sounds good. 

John

PR #22, opened 2018-12-15
Phase1 update for external inclusion

hjmjohnson:


hjmjohnson:
@leej3 
I've completed the refactoring!  It took a bit longer than expected, but I ended up doing a lot more than originally expected.  

I also fixed the testing where running tests in parallel would randomly fail based on the order of running order.  

I also integrated as an external copy into ITK and verified that the ITK build works correctly.  I think this same approach will work for AFNI as well!.

hjmjohnson:
@afni-rickr Rick.  I hope to merge this on Sunday afternoon. It includes greatly improved cmake support for nifti.   You might want to review (retrospectively) to see how to integrate this with 
AFNI from the demo I put together here:  https://github.com/NIFTI-Imaging/nifti_clib/blob/phase1-update-for-external-inclusion/real_easy/Demonstrations.md

leej3:
Fantastic. Quite the refactor. A few comments/questions:

```
+ In the cifti cmake file, find_package for expat does not work for me. include(FindEXPAT) works nicely. Is there a reason not to use that?

+ I don't quite understand why the target_include_directories use INTERFACE rather than PUBLIC. Is this to avoid building targets with source files one hasn't explicitly specified?

+ There's a couple of times where a sources variable is defined. Can we contract this simply to add_x(target_name source_x.c). This seems cleaner.

+ The test data fetching is nice. Is this cached on travis? I know there's a caching mechanism on circleci. I know that @afni-rickr  has many more tests that could be added if we could happily add more data to the process.

+ Should travis be using 3.11 since that is the minimum version specified?

+ Currently dynamic linked installs don't work for me with a simple make install. I know very little about cpack, which seems how one would distribute in general and so maybe this is not required. I know that it requires messing around with rpath. Is that a phase 2 thing? I'm just wondering, it seems that CPACK will cover most mechanisms for distribution. I have also not fully digested all of the examples of integration for downstreams. Perhaps I will find my answer there (I won't get to that today).

+ As an aside, I just got the Cmake cookbook. It's a wonderful synthesis of all the ideas I have been scraping from all the blogposts (and attempting to synthesize myself). 
```

hjmjohnson:
@leej3 All good points.  I'll address each of those today.  I have graduation today from 10-1 so I don't think I'll get this done before then.

Thank you for the careful review.

*  I may need your help with the Expat configuring.  I always prefer find_package paradigm, and I don't know why include(FindEXPAT)  works over find_package().  The find_package worked on my mac.  I'll build on Linux and determine if the include(FindEXPAT) works.
* NICE CATCH!  These should be PUBLIC.   It was a copy-n-paste bug.
* I'll remove the extra variables.  I had it on my todo list but left that list on my desk as work.
* I created a separate repo for testing data.  The repo is:  https://github.com/NIFTI-Imaging/nifti-test-data if data is added there, just make a new release (which makes a new tarball automatically) and then update the link locally.  You have been invited to that repo.
* I don't think travis needs to be at the minimum cmake version.  Ideally, the version used by travis would be all possible environment configurations, but that is unreasonable to accomplish.
* CPACK is definitely a phase 2 effort.
* Thanks.  I'm ordering the cmake book on monday.

hjmjohnson:
@leej3 I'm merging this set of patches, and will address your concerns as separate items later today.  If you have a moment to make a correction that works for you regarding the EXPAT, I'd greatly appreciate that.

afni-rickr:
Hi @hjmjohnson, that is a big update.  Did you re-write the test scripts under 'commands' (or maybe just some of them) in bash, is that what cmake_testscripts is?  I guess that is the build test, rather than a regression test then, is that right?  Thanks!

hjmjohnson:
@afni-rickr looking back through the logs:

1)  Changed permission for files that start with "#!/bin/*" to be `chmod +x `
2) Removed blank line at end of file (Perhaps that was a mistake)

Is there something that needs to be reverted?  All o the cmake tests are passing, so I assumed this was working properly. 

Hans

hjmjohnson:
@afni-rickr FYI  cmake_testscripts was moved, but it has existed for a long time.  I updated the cmake that ran those scripts, but I did not modify those scripts in any substantive way.

It appears that I wrote those scripts back in 2007

`git log --follow nifti2/nifti_regress_test/cmake_testscripts/bricks_test.sh`

afni-rickr:
Oh, I didn't even realize those files under cmake_testscripts existed (for 11 years now)!  Sorry, I thought you had just added them.  They are based on the tcsh command scripts under nifti_regress_test/commands.  I don't know who actually contributed them.

afni-rickr:
Oh, great!  In any case, that ctest is indeed working for me, I was just thinking it was new.

afni-rickr:
Hi @hjmjohnson. The build-stand-alone cmake command is not working for me.  Here is what that cmake command shows:
`cmake -DNIFTI_DIR:PATH=../build-nifti_clib ${STAND_ALONE_SOURCE_DIR}

CMake Error at CMakeLists.txt:13 (find_package):
  Could not find a package configuration file provided by "NIFTI" (requested
  version 2.1.0) with any of the following names:
    NIFTIConfig.cmake
    nifti-config.cmake
  Add the installation prefix of "NIFTI" to CMAKE_PREFIX_PATH or set
  "NIFTI_DIR" to a directory containing one of the above files.  If "NIFTI"
  provides a separate development package or SDK, be sure it has been
  installed.
`
Is there anything you would like me to try out?  Thanks.

PR #23, opened 2018-12-16
COMP: The target_include_directories should be PUBLIC

hjmjohnson:
Include directories associated with the library targets
should be propagated to the external targets that use them
with PUBLIC behavior.

Co-authored-by: leej3 <johnleenimh@gmail.com>


leej3:
looks good

PR #29, opened 2018-12-16
Fix type versions

hjmjohnson:


hjmjohnson:
YEAH.  Back to a green dashboard.

PR #34, opened 2018-12-17
STYLE: Remove the register keyword

hjmjohnson:
The register keyword is a hint to the compiler that the variable will be
heavily used and that you recommend it be kept in a processor register
if possible.

Most modern compilers do that automatically, and are better at picking
them than us humans.


afni-rickr:
Yes, it is nice to get those out, thanks.

Issue #37, opened 2018-12-18
Link Time Optimization (LTO) fails due to duplicate symbols


hjmjohnson:
@afni-rickr I'd love your advice/help in identifying a way to resolve this linkage problem.

#31 builds identify problems that will prevent LTO (inter procedure optimization).

When building nifti2, these symbols violate the "One Definition Rule (ODR)"

ld: 98 duplicate symbols for architecture x86_64
duplicate symbol _nifti_disp_matrix_orient in:
    libnifti2.a(nifti2_io.c.o)  ../niftilib/libniftiio.a(nifti1_io.c.o)
duplicate symbol _nifti_image_load_bricks in:
    libnifti2.a(nifti2_io.c.o)    ../niftilib/libniftiio.a(nifti1_io.c.o)
duplicate symbol _nifti_image_free in:
    libnifti2.a(nifti2_io.c.o)    ../niftilib/libniftiio.a(nifti1_io.c.o)
duplicate symbol _nifti_update_dims_from_array in:
    libnifti2.a(nifti2_io.c.o)    ../niftilib/libniftiio.a(nifti1_io.c.o)
duplicate symbol _valid_nifti_brick_list in:
    libnifti2.a(nifti2_io.c.o)    ../niftilib/libniftiio.a(nifti1_io.c.o)
duplicate symbol _nifti_intent_string in:
    libnifti2.a(nifti2_io.c.o)    ../niftilib/libniftiio.a(nifti1_io.c.o)
duplicate symbol _nifti_slice_string in:
    libnifti2.a(nifti2_io.c.o)    ../niftilib/libniftiio.a(nifti1_io.c.o)
duplicate symbol _nifti_datatype_sizes in:
    libnifti2.a(nifti2_io.c.o)    ../niftilib/libniftiio.a(nifti1_io.c.o)
duplicate symbol _nifti_quatern_to_mat44 in:
    libnifti2.a(nifti2_io.c.o)    ../niftilib/libniftiio.a(nifti1_io.c.o)
duplicate symbol _nifti_mat44_to_orientation in:
    libnifti2.a(nifti2_io.c.o)    ../niftilib/libniftiio.a(nifti1_io.c.o)
duplicate symbol _nifti_orientation_string in:
    libnifti2.a(nifti2_io.c.o)    ../niftilib/libniftiio.a(nifti1_io.c.o)
duplicate symbol _nifti_datatype_string in:
[...]

PR #38, opened 2018-12-18
ENH: Decouple nifti1 and nifti2 libraries

hjmjohnson:
Make a duplicate copy of the nift1.h header
in the nifti2 directory in order to decouple
the dependencies between these two libraries.

Fixes: #37

### DIFF
--- a/nifti2/CMakeLists.txt
+++ b/nifti2/CMakeLists.txt
@@ -1,7 +1,7 @@
 set(NIFTI_NIFTILIB2_NAME ${NIFTI_PACKAGE_PREFIX}nifti2)
 
 add_library(${NIFTI_NIFTILIB2_NAME} nifti2_io.c )
-target_link_libraries( ${NIFTI_NIFTILIB2_NAME} PUBLIC ${NIFTI_PACKAGE_PREFIX}niftiio)
+target_link_libraries( ${NIFTI_NIFTILIB2_NAME} PUBLIC ${NIFTI_PACKAGE_PREFIX}znz)
 target_include_directories(${NIFTI_NIFTILIB2_NAME} PUBLIC
         $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>
         $<INSTALL_INTERFACE:${NIFTI_INSTALL_INCLUDE_DIR}>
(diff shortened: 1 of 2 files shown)


nifti2/CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 set(NIFTI_NIFTILIB2_NAME ${NIFTI_PACKAGE_PREFIX}nifti2)
 
 add_library(${NIFTI_NIFTILIB2_NAME} nifti2_io.c )
-target_link_libraries( ${NIFTI_NIFTILIB2_NAME} PUBLIC ${NIFTI_PACKAGE_PREFIX}niftiio)
+target_link_libraries( ${NIFTI_NIFTILIB2_NAME} PUBLIC ${NIFTI_PACKAGE_PREFIX}znz)
```
leej3: @hjmjohnson out of interest what's the easiest way to work this out? The transitive dependency of niftiio prevents failure but is obviously inappropriate to use. Is there a tool that you use to work out the appropriate targets to link to or do you just start from nothing and start adding the correct targets that are required for linking?

hjmjohnson: @leej3 What failure are you referencing?  I'm not sure what you are asking for.

leej3: In this context, the question doesn't quite make sense. The goal of these changes was to remove niftiio as a dependency. 
Ignoring that context though. If nifti2 was dependent on znz and was failing to link, this could be resolved by linking to niftiio (since it exposes znz during the linking process anyway). The linking would work but the dependency would not really be correctly specified (or at least that is what I believe to be the case). My question was what was your process for determining what to link against. Do you just check the C code, start by linking nothing and work up from there, or is there a tool you use?

leej3:
Am I correct in that it will only start duplicated. As nifti2 evolves the
header files will diverge right? Alternatively, in cmake we could just add
the nifti1 header to an interface library and add it as a PRIVATE include
for nifti2. Does this solve the duplicate symbols issue though. @hjmjohnson,
the current solution was best because it is the only way to adequately
decouple the two libraries correct?

leej3:
> @leej3 Ahh. Well sorry, but there is no easy tool for determining the "minimum linkage path".
> 
> You might want to look at #37 to see how to make plots of dependencies that you have already created and review that to see if there are optimizations.

I saw that alright. A nice overview of what's happening. Something I will be using. I also found some useful tools [here](https://cliutils.gitlab.io/modern-cmake/chapters/features/utilities.html). The linking was feeling painfully iterative and error fraught, so I had hoped for a magical tool.

hjmjohnson:
The problem with the link libraries occurs due to nifti2_io.c and nifti_io.c defining the same symbols.  It is required that only one of libniftiio or libnifti2 be linked to the same binary.

Hans

hjmjohnson:
The previous problem was the the nifti1.h header was bound exclusively to the libnifti2 library, so now there are 2 identical versions of nifti1.h one bound to the libniftiio and one bound to libnifti2.

PR #39, opened 2018-12-18
Fix scanbuild issues

hjmjohnson:


leej3:
@hjmjohnson, thank you. It's an instructive gathering of all of the changes. I am not yet proficient in C so I certainly cannot comment there, though I have noted the various lurking bugs and shall attempt to exterminate them elsewhere.

Did you use cmake or travis documentation to reconfigure the testing? I'm just wondering where the most useful guides for this are.

Regarding the linking to libm. They pick up this dependency from znz. Is this simply that they independently require the math library so one should specify it anyway?

hjmjohnson:
@leej3 The missing function error that required libm was "sqrt" which only exists in 

```
git grep -l sqrt
nifti2/nifti1.h
nifti2/nifti2_io.c
nifticdf/nifticdf.c
niftilib/nifti1.h
niftilib/nifti1_io.c
```

That is why I put the dependencies in each of those libs separately.

leej3:
ok. noted for the future. thaks

PR #47, opened 2018-12-20
Second pass scanbuild fixes

hjmjohnson:


cifti/afni_xml.c
```diff
@@ -145,50 +144,52 @@ afni_xml_list axml_read_file(const char * fname, int read_data)
...
       return xlist;
 
-   xd->xroot = &xlist;      /* return struct */
-   xd->dstore = read_data;  /* store for global access */
+   gAXD.xroot = &xlist;      /* return struct */
+   gAXD.dstore = read_data;  /* store for global access */
 
    if( !fname ) {
       fprintf(stderr,"** axml_read_image: missing filename\n");
+      gAXD.xroot = NULL;
```
hjmjohnson: @afni-rickr The aliasing of gAXD by xd  made the code really hard to understand with respect to the error.  Changing from "xd->" to "gAXD" and "xd" -> "&gAXD" allowed me to identify the real problem.

The issue is that `gAXD.xroot` holds the address of the local variable "xlist", which is OK within the scope of this function, but then when the function ends gAXD.xroot will have an address to a variable that no longer exists.

The real fix was adding lines 155, 162, 168, and 207 to ensure that when this function ends, the global variable gAXD no longer has any references to local variables.

I found it really hard to find the bug because xd looked like a local variable at first glance, but it is really just an alias.

If the aliasing is a coding style you prefer, I can re-alias the global variable.

hjmjohnson: I don't agree with the statement "avoid lots of references to the global structure".  There are still references to the global structure, they are just named differently.

Was the intent to "copy" the global structure as an initializer?

hjmjohnson: @afni-rickr Do you have any further suggestions?  I'd like to complete this PR so that the next set of work that is stacking up on top of this can be ordered for inclusion.

hjmjohnson: @afni-rickr Once we get this PR to the point of your acceptance, there will be 2-3 more PR's that greatly improve the testing environment.

afni-rickr: The realloc change is good, I did not realize the old memory would be preserved, or that free would be universally happy with a NULL pointer.
In afni_xml.c, resetting xroot to NULL is nice, though unnecessary.  I probably did not bother since xroot is only used in the context of these functions.  Some of the frees are not necessary, but they are fine.
So I guess this is all okay, even just to appease scanbuild (is that was is reporting them?), though I do prefer the aliasing with xd.  I am sorry going through this took time.  Feel free to pass warnings about my code off to me, too.

hjmjohnson: Thanks @afni-rickr .  I will revert the code to using the "xd" and make a new pull request in a few minutes.

PR #51, opened 2018-12-21
fix bound check and pointer inc

afni-rickr:


afni-rickr:
@hjmjohnson Quick fix before running off.  The first is for the new bound check.  The second I just happened to see when looking at the first.

afni-rickr:
@hjmjohnson Thanks, you too!

PR #52, opened 2018-12-26
Cifti

afni-rickr:
comments and wrap-around


hjmjohnson:
@afni-rickr I rebased the commit on the master branch to simplify it's history.  Originally this was a patch on top of a merge.  That history was unnecessarily more complex than needed.

I personally prefer avoiding merges.  `git fetch origin; git rebase origin/master` keep a much cleaner, linear history with each patch representing one easy to understand change.

afni-rickr:
Thanks @hjmjohnson.  I noticed the history and was going to delete and re-create the branch.  This is nicer.

PR #58, opened 2019-01-04
NIFTI-2 udpates

afni-rickr:
- fix extension offset for NIFTI-2
- add -mod_hdr2 option to nifti_tool


afni-rickr:
@hjmjohnson Actually, the features are due to the testing that I am writing.  Putting them into the context of cmake is more work, particularly in that much of the testing is iterative.  But maybe I could add a couple of atomic tests and we can ponder being more elaborate.  Thanks.

PR #61, opened 2019-01-11
Adjust scan-build invocation to cause travis to fail on issues found

gdevenyi:
Make the scan-build invocation to return error if anything is found, so that travis tests don't pass if a PR introduces a new (possible) bug.


gdevenyi:
Looks like this is working, except I think ctest is suppressing the actual "make" output, so if scan-build throws an error it might be a bit hard to track down.

I don't know ctest at all to attempt to expose that output so one could see it...

PR #62, opened 2019-01-18
Update installation

leej3:
@hjmjohnson  as described in [gifts_clib](https://github.com/NIFTI-Imaging/gifti_clib/pull/4) I had some problems attempting to work with the nifti_clib installation as is. I have now failed around 20 tests that I shall have to work through. Otherwise the improvements will make the code easier to maintain and downstreams will be able to incorporate it into their builds more easily.


CMakeLists.txt
```diff
@@ -174,7 +174,7 @@ if(NIFTI_INSTALL_EXPORT_NAME EQUAL "NIFTITargets")
 
   export(EXPORT NIFTITargets
         FILE "${CMAKE_CURRENT_BINARY_DIR}/NIFTI/NIFTITargets.cmake"
-        NAMESPACE ${NIFTI_PACKAGE_PREFIX}NIFTI::
```
hjmjohnson: This is needed for incorporating NIFTI in other packages natively to avoid namespace collisions.  I really do not want to have to manually add this back in for ITK.

CMakeLists.txt
```diff
@@ -145,37 +146,36 @@ if(USE_FSL_CODE)
...
 
-if(NIFTI_INSTALL_EXPORT_NAME STREQUAL "NIFTITargets")
   #######################################################################
   # CMake itself and can use some CMake facilities for creating the package files.
   # This allows for find_package(NIFTI 2.1.0 NO_MODULE) to work for pulling in
   # NIFTI libraries into an external project
   include(CMakePackageConfigHelpers)
-  set(CONFIG_SETUP_DIR ${CMAKE_CURRENT_BINARY_DIR}/NIFTI)
-  set(ConfigPackageLocation share/cmake/NIFTI)
+  set(CONFIG_SETUP_DIR ${CMAKE_CURRENT_BINARY_DIR}/${PACKAGE_NAME})
```
hjmjohnson: Nice!

cifti/CMakeLists.txt
```diff
@@ -19,20 +20,14 @@ if(BUILD_SHARED_LIBS)
     set_target_properties(cifti PROPERTIES ${NIFTI_LIBRARY_PROPERTIES})
 endif()
 
-if(NOT NIFTI_INSTALL_NO_LIBRARIES)
```
hjmjohnson: Why is this removed?

fsliolib/CMakeLists.txt
```diff
@@ -10,23 +11,15 @@ if(BUILD_SHARED_LIBS)
   set_target_properties(fslio PROPERTIES ${NIFTI_LIBRARY_PROPERTIES})
 endif()
 
-if(NOT NIFTI_INSTALL_NO_LIBRARIES)
```
hjmjohnson: why is this removed?

fsliolib/CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
 add_library(fslio fslio.c )
-set_target_properties(fslio PROPERTIES PUBLIC_HEADER fslio.h)
+set_target_properties(fslio PROPERTIES PUBLIC_HEADER "fslio.h;dbh.h")
```
hjmjohnson: nice.

CMakeLists.txt
```diff
@@ -76,9 +76,9 @@ endif()
 set_if_not_defined(NIFTI_PACKAGE_PREFIX "")
 set(PACKAGE_NAME ${NIFTI_PACKAGE_PREFIX}NIFTI)
 set(NIFTI_INSTALL_EXPORT_NAME ${PACKAGE_NAME}Targets)
-set_if_not_defined(NIFTI_INSTALL_RUNTIME_DIR "${CMAKE_INSTALL_PREFIX}/bin")
-set_if_not_defined(NIFTI_INSTALL_LIBRARY_DIR "${CMAKE_INSTALL_PREFIX}/lib")
-set_if_not_defined(NIFTI_INSTALL_INCLUDE_DIR "${CMAKE_INSTALL_PREFIX}/include")
+set_if_not_defined(NIFTI_INSTALL_RUNTIME_DIR "${CMAKE_INSTALL_PREFIX}/bin/${PACKAGE_NAME}")
```
hjmjohnson: This seems like a very bad idea.  Can you elaborate on the problem encountered?

Also "//' should be "/"

leej3:
Thanks for the feedback!

> There is a lot going on here. I'm in the middle of teaching 2 courses and writing a grant so I can not do any testing.
> 
> I'm 85% sure that this is going to totally break all possibility of integrating directly with ITK. When building NIFTI for ITK use, we need the libraries to be called "itkniftilib.a" and similar to avoid namespace collisions.

There's no rush. I'm just working on passing the tests and I'll try to add some. And then I have to figure out CPACK so that I can add it to gifti_clib.

I accidentally dropped the lib for libraries but have remedied that now and the behavior should be as desired: with the prefix set to "itk" the nifti2 target will be installed as libitknifti2.a (or libitknifti2.so) and imported as itkNIFTI::itknifti2. Let me know if I am still missing something. Assuming we move the macros into a module this version of the build removes a lot of the NIFTI_PACKAGE_PREFIX variables in the files. I'm happy to revert if you think its unnecessary or convoluted, I just wanted to copy from the cmake-cookbook to get to a version of the build that I could use find_package with. Now that I'm in that state I can make changes along the way. Also, I have no attachment to the various defaults I have set. I have attempted to maintain the same behavior: change as you deem appropriate.

 I'm assuming that here by namespace collisions you mean that if a nifti install and itk-nifti install exist side by side the filenames will collide. (as in to avert it the above needs to be done). If that's not correct let me know.

CMakeLists.txt
leej3: I think I have fixed this now

fsliolib/CMakeLists.txt
leej3: I was trying to drop the number of variables to define. The COMPONENTS option in the install command allows a user to install their choice of executables,libraries, and header files. Happy to change this back if this wreaks havoc elsewhere though

CMakeLists.txt
leej3: Described below, I think I have now fixed this

hjmjohnson:
@leej3 Re-open if you attempt to address my previous comments.

leej3:
@hjmjohnson, if you have some time could you have another look at this now. I've tidied up everything and made sure it works with the ITK build (passes all tests).

leej3:
Sigh, once I've passed the tests at least. Some failures in this repository.

hjmjohnson:
https://my.cdash.org/viewConfigure.php?buildid=1808112

See the errors.

I think it has to do with fslio

INSTALL TARGETS - target fslio has PRIVATE_HEADER files but no PRIVATE_HEADER DESTINATION.
Installed_targets: znz;niftiio;nifti1_tool;nifticdf;nifti_stats;nifti2;nifti_tool;cifti;afni_xml_tool;cifti_tool
CMake Error at CMakeLists.txt:210 (install):
  install TARGETS given target "znz" which does not exist in this directory.

leej3:
Thanks. I'm puzzled why the bug is only showing on travis. The docker image builds happily enough (it builds the fsl stuff).

I'll work it out. I may run the travis stuff locally to make the debugging a little more tractable.

hjmjohnson:
@leej3 are you doing "make install" step?

leej3:
I wasn't using "make install" in the dockerfile (and should) but that didn't seem to fix it. I'm working through the details in the travis docker image now:
```
BUILDID="build-$RANDOM"
INSTANCE="travisci/ci-ubuntu-1804:packer-1582574251-a9b1ae58"
docker run --name $BUILDID -dit $INSTANCE /sbin/init
docker exec -it $BUILDID bash -l
```

leej3:
Well the error reproduces, which is reassuring. And it is indeed the install step failing. I think it's a scoping issue based on the hack that if I wrap the install targets in quotes I get:
```
CMake Error at CMakeLists.txt:210 (install):
  install TARGETS given target
  "znz;niftiio;nifti1_tool;nifticdf;nifti_stats;nifti2;nifti_tool;cifti;afni_xml_tool;cifti_tool"
  which does not exist in this directory.

 I'm presuming target scoping is not as global in earlier versions of Cmake or something like that. The earliest version of cmake I was using was 3.13.4. And most of the time I was on 3.16.5. Assuming it's that it shouldn't be too hard to fix. Should get it sorted early tomorrow.

**Comment by hjmjohnson** [comment]
I think we should target at least back to 3.10.2 for cmake support.  Perhaps even earlier. 

3.10.2 is the default in Ubuntu 18.04.

Hans

**Comment by leej3** [comment]
Agreed. 

John

**Comment by gdevenyi** [comment]
Probably the version on 16.04, or maybe whatever is in centos7

**Comment by leej3** [comment]
Well this is a little messier but it works (on 18 LTS, cmake 3.10).

The main problem is indeed that until 3.13 targets do not exist beyond the directory scope. So the install command needs to occur in that specific directory. To at least consolidate the installation logic I have encapsulated it in a function. 

Unfortunately this means that the exported targets are not available  in the project directory. I'm not sure if there is a solution for this. For now I have just disabled config/version/target installation for older cmake versions. 
I believe these changes still work with ITK (which already handles the process of target export and doesn't write out  the config/version modules that I can see) but I will test that now and then tidy all of this up.

**Comment by leej3** [comment]
@gdevenyi centos7 is 2.8 I think and 16.04 is 3.5. I struggle a little with 3.10 so I don't think I could manage a version that early. The min version set before this pull request was 3.10 though. There are already  some issues on 3.5 to work through if we do want to push for supporting 16.04.

File: Dockerfile [hunk 7]
```diff
@@ -10,19 +12,38 @@ RUN apt-get update && apt-get install -y -q eatmydata \
         zlib1g-dev \
     && rm -rf /var/lib/apt/lists/*
 
+# Install newer cmake version with globally scoped targets
+ENV CMAKE_VER=3.13.3
```
hjmjohnson, line 16: This should be 3.10.2 if that is the version that we are targeting for support.

Dockerfile
```diff
@@ -10,19 +12,38 @@ RUN apt-get update && apt-get install -y -q eatmydata \
...
+#     -DDOWNLOAD_TEST_DATA=OFF \
+#     /opt/src/nifti_clib \
+#     && make install \
+#     && ctest --output-on-failure -LE NEEDS_DATA
+
+ENV PATH="/opt/cmake/cmake-${CMAKE_VER}-Linux-x86_64/bin:$PATH"
 RUN cmake \
-    -DBUILD_SHARED_LIBS=ON \
-    -DNIFTI_USE_PACKAGING=ON \
+    -DBUILD_SHARED_LIBS=OFF \
```
hjmjohnson, line 42: Shared libraries are typically more difficult to build.  Whey turn off packaging?

gdevenyi:
Personally, I'm fine with a higher minimum version, just be aware that you're likely to get support requests when this won't build on these long-term support distributions.

leej3:
> @leej3 Are you satisfied with this build system? Has it been tested with ITK?

Yes it worked for me with ITK. I would love to figure out how to support the installation of the export set on 3.10, then that all of the build system's functionality will be supported back to 3.10 as far as I can tell (though the tests with the NEEDS_DATA label will still require 3.11). 

I'll try to figure out the export set stuff today. My understanding is that using the EXPORT option of the install command writes out the export set to file. Given 3.10 does not all targets to be installed from the same directory I'm not sure how to include them all in the same export set. If you have any pointers they would be welcome.

Otherwise I am happy with the pull request. I think it will make things a little easier to maintain and that the additional support for downstreams will be useful.

Dockerfile
leej3: The fact that it was missing the export set stopped me from using it. Agreed though, I will change this to 3.10 instead.

Dockerfile
leej3: Noted. Will change this.

leej3:
> Personally, I'm fine with a higher minimum version, just be aware that you're likely to get support requests when this won't build on these long-term support distributions.

Noted. I think I am willing to relax this for gifti but it would be great to hit 3.10 for nifti, it's used by many projects and will hopefully used by many more.

leej3:
Ok I'm happy with this now. All build and install functionality is supported back to 3.10. 

Once support is a minimum of 3.13 the additional logic implemented in `install_nifti_target` and the additional calls of that function within the target specific CMakeLists.txt files could be removed to tidy things up.

 I will looking into back porting the test data download functionality to 3.10 in a separate pull request.

Dockerfile
hjmjohnson: "If it's not tested, it's broken".  Either remove it or test it.

hjmjohnson:
> Ok I'm happy with this now. All build and install functionality is supported back to 3.10.
> 
> Once support is a minimum of 3.13 the additional logic implemented in `install_nifti_target` and the additional calls of that function within the target specific CMakeLists.txt files could be removed to tidy things up.
> 
> I will looking into back porting the test data download functionality to 3.10 in a separate pull request.

I would not backport testing.  Any developer should be using 3.13.  The use of 3.10 is only to support integration with package managers like neuro-debian, or integration with existing toolkits like ITK.

Dockerfile
leej3: Reasonable. So you don't know of anyone who uses the packaging. I know [we proposed supporting it](https://docs.google.com/document/d/1tZLNs_HEFFRxZ8Cj1_f3ddMV2tc6OmxWKMTCZK8WV5s/edit#heading=h.8v3b0htx4dtu) but I think that was me. And I would be happy with debian, an rpm distribution, macOS (likely via home-brew) and Conda for releases. After that the tar files from releases on GitHub seems adequate. In which case it can indeed be removed.

leej3:
> > Ok I'm happy with this now. All build and install functionality is supported back to 3.10.
> > Once support is a minimum of 3.13 the additional logic implemented in `install_nifti_target` and the additional calls of that function within the target specific CMakeLists.txt files could be removed to tidy things up.
> > I will looking into back porting the test data download functionality to 3.10 in a separate pull request.
> 
> I would not backport testing. Any developer should be using 3.13. The use of 3.10 is only to support integration with package managers like neuro-debian, or integration with existing toolkits like ITK.

Aha. Good to know. Thanks

PR #63, opened 2019-03-11
NIFTI_XFORM_TEMPLATE_OTHER

afni-rickr:


seanm:
@afni-rickr After investigating something unrelated, I just noticed that `NIFTI1_XFORM_CODES` got a new possibility `NIFTI_XFORM_TEMPLATE_OTHER` and I traced it back to this PR / commit.

I don't see a lot of info here about what this is meant to represent... Could you elaborate?

Also, are there any public example .nii files using this code?  I'd like to add one to my app's own unit tests.

Thanks.

afni-rickr:
Hi @seanm .  There is a code for the Talairach template (3) and one for MNI (4), but that was it (for templates).  But there are many possible standard space templates (maybe even 10 or so just for "MNI" space, which is not unique even as a general space).  The ALIGNED_ANAT is typically an original space alignment.
So it begged for a code that implied a template space, but not TLRC or MNI.  Does that seem reasonable?

seanm:
@afni-rickr yes, thanks for the explanation.

I guess I can just hack an existing file's header to create such a file...

seanm:
Oooh, nice.  I was just going to use a hex editor. :)

PR #67, opened 2019-05-12
ENH: Add intent codes used by FSL to avoid future clashes

### LINKED ISSUE #66: Reserve or codify FSL intent codes
Just want to flag that FSL has custom intent codes for [FNIRT](https://github.com/mshvartsman/FSL/blob/7aa2932949129f5c61af912ea677d4dbda843895/src/warpfns/fnirt_file_reader.h#L100-L107) and [TOPUP](https://github.com/mshvartsman/FSL/blob/fsl-5.0.9/src/topup/topup_file_io.h#L100-L106) files.

Because these exist in the wild, we accept them in nibabel (nipy/nibabel#528; [nibabel/nifti1.py#L239-L251](https://github.com/nipy/nibabel/blob/ed3cdb5a33fdf1dd49e037600927f3b213f61f08/nibabel/nifti1.py#L239-L251)).

I don't have a strong opinion on whether these should be adopted in the official standard, but they should probably be reserved with a note, so a clash doesn't occur in the future.

cc @pauldmccarthy

effigies:
These intent codes were present in the FSL 5.0.9 source code. I have not downloaded more recent versions to see if they have added new codes.

@pauldmccarthy I figure someone from the FSL world should probably have a look over this.

Closes #66.

### DIFF
--- a/nifti2/nifti1.h
+++ b/nifti2/nifti1.h
@@ -957,6 +957,33 @@ typedef struct { unsigned char r,g,b; } rgb_byte ;
 
 #define NIFTI_INTENT_SHAPE        2005
 
+ /*! The following intent codes have been used by FSL FNIRT for
+     displacement/coefficient files.
+
+     These codes are included to prevent clashes in community-created
+     extensions to NIfTI. Encoding and decoding behavior for these
+     intents is not specified by the standard, and support is OPTIONAL
+     for conforming implementations.
+     */
+
+#define NIFTI_INTENT_FSL_FNIRT_DISPLACEMENT_FIELD       2006
+#define NIFTI_INTENT_FSL_CUBIC_SPLINE_COEFFICIENTS      2007
+#define NIFTI_INTENT_FSL_DCT_COEFFICIENTS               2008
+#define NIFTI_INTENT_FSL_QUADRATIC_SPLINE_COEFFICIENTS  2009
+
+ /*! The following intent codes have been used by FSL TOPUP for
+     displacement/coefficient files.
+
+     These codes are included to prevent clashes in community-created
+     extensions to NIfTI. Encoding and decoding behavior for these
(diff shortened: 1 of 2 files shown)


pauldmccarthy:
Hey @effigies, sorry for the delay - these are all the FSL-specific codes that I am aware of. 

The  `_COEFFICIENTS` file types also do non-standard things with the sform, qform, and pixdim fields (to store extra information needed to reconstruct a full displacement field from the coefficients), but I'm not sure if that's something that this library (or `nibabel`) should be responsible for handling?

hjmjohnson:
@effigies @pauldmccarthy I would prefer to have the codes listed in the standard with a comment that implementation of encode/decode of images with this intent is OPTIONAL, and not explicitly supported by the standard.

nifti2/nifti1.h
```diff
@@ -957,6 +957,21 @@ typedef struct { unsigned char r,g,b; } rgb_byte ;
 
 #define NIFTI_INTENT_SHAPE        2005
 
+ /*! The following intent codes have been used by FSL FNIRT for
+     displacement/coefficient files.  */
```
hjmjohnson: @effigies Discussion that these intent flags support community supported extensions to NIFTI, but implementing the encode/decode behaviors of these intents is not supported by the NIFTI standard.

effigies:
@hjmjohnson Thanks for the review. Added the following boilerplate above each of the two groups of intents:

```
     These codes are included to prevent clashes in community-created
     extensions to NIfTI. Encoding and decoding behavior for these
     intents is not specified by the standard, and support is OPTIONAL
     for conforming implementations.
```

Let me know if you prefer other wording/organization.

effigies:
@pauldmccarthy I would say that it's mostly the responsibility of the library to provide reliable representations of the encoded structures so that the application can choose how to interpret them.

PR #74, opened 2019-09-25
remove dupe close

afni-rickr:


nifti2/nifti2_io.c
```diff
@@ -5344,7 +5344,6 @@ nifti_2_header * nifti_read_n2_hdr(const char * hname, int * swapped,
       if( g_opts.debug > 1 )
          fprintf(stderr,"++ reading ASCII header via NIFTI-2 in %s\n", hname);
       nim = nifti_read_ascii_image(fp, hname, -1, 0);
-      znzclose(fp) ;
```
hjmjohnson: @afni-rickr I don't think that nifti_read_ascii_image should close the fp.  I think that the block of code that opens the file should also close the file.  

If you disagree, then nifti_read_ascii_image should have it's documentation updated to state that a side effect of running that function is that the fp is closed.

afni-rickr: @hjmjohnson  Either way makes sense at this point.  Looking back, it was a terminal replacement branch for/from nifti_image_read for this case.  However given that, it might have been more appropriate to not export the function in the first place.  Now, not closing is a functional change on 15 year old code.
I will code up a change and see what it looks like.

PR #77, opened 2019-09-30
Cov

afni-rickr:
this is a partial addition of some coverage tests


afni-rickr:
** removing: this branch doesn't have previous updates to nifti?_io.c..., sorry

PR #80, opened 2019-10-07
Nt tests

afni-rickr:
This should add more coverage testing, mostly via nifti_tool.  The updates to the NIFTI library itself are minimal, basically adding a couple of matrix multiplication functions.

This includes all of the testing additions that I want for now.


afni-rickr:
@hjmjohnson For the record, the review request is maybe just to see that the structure of the testing seems reasonable.  I would not ask you to read through all of this...  :)

afni-rickr:
@hjmjohnson It is not clear to my what these failures are from.  Can you tell?

PR #81, opened 2019-10-07
ENH: Update linux default builds to bionic 18.04.

hjmjohnson:


hjmjohnson:
@afni-rickr FYI:  I was able to fix the cdash reports at
 
https://my.cdash.org/index.php?project=nifti_clib

PR #82, opened 2019-10-07
Increase coverage of znzlib

hjmjohnson:


hjmjohnson:
Addressing low hanging fruit from:
https://my.cdash.org/viewCoverageFile.php?buildid=1711875&fileid=3622840

https://my.cdash.org/viewCoverage.php?buildid=1711875&status=6&dir=znzlib

PR #83, opened 2019-10-07
Fix compiler warnings

hjmjohnson:


hjmjohnson:
@afni-rickr 

https://my.cdash.org/index.php?project=nifti_clib

With hash `3e4d38`

afni-rickr:
This is starting to go a bit far, perhaps.  I would prefer not to make too many style changes.  Are these all coming from cdash?

afni-rickr:
I am sorry, but it isn't clear to me what to review regarding that 3e4d38 hash...  If it is the warnings at the top, I do not find them important, except to appease a whiny compiler. :)

hjmjohnson:
> This is starting to go a bit far, perhaps. I would prefer not to make too many style changes. Are these all coming from cdash?

My experience is that these style changes will help to foster a community that can assist with supporting nifti.  I know that you are the primary developer, but at some point, it would be nice if others were enticed to help support the toolkit.

PR #86, opened 2019-10-08
Minor clang tidy warnings

hjmjohnson:


hjmjohnson:
Breaking up huge patch set #83 into smaller units.

> Comments from @afni-rickr
> 
> If you break it up, the one commit I would like left out is 9b3b8b4. If it would be helpful, I can resolve that neglected errs var in the nt_tests branch. Well, should nt_tests be merged for that matter? I was not sure whether there was an interaction with this pull request. Thanks!
> 
> Otherwise, I have looked through all of these commits and do not see any problems with them.

PR #87, opened 2019-10-08
WIP: DO NOT COMMIT -- This is only here to identify work to be done.

hjmjohnson:
COMP: Remove 'value never used' warnings.

### DIFF
--- a/niftilib/nifti1_io.c
+++ b/niftilib/nifti1_io.c
@@ -3180,21 +3180,18 @@ int nifti_type_and_names_match( nifti_image * nim, int show_warn )
             fprintf(stderr,
             "-d NIFTI_FTYPE 1, but no .nii extension in header filename, %s\n",
             nim->fname);
-         errs++;
       }
       if( fileext_n_compare(ext_i,".nii",4) ) {
          if( show_warn )
             fprintf(stderr,
             "-d NIFTI_FTYPE 1, but no .nii extension in image filename, %s\n",
             nim->iname);
-         errs++;
       }
       if( strcmp(nim->fname, nim->iname) != 0 ){
          if( show_warn )
             fprintf(stderr,
             "-d NIFTI_FTYPE 1, but header and image filenames differ: %s, %s\n",
             nim->fname, nim->iname);
-         errs++;
       }
    }
    else if( (nim->nifti_type == NIFTI_FTYPE_NIFTI1_2) || /* .hdr/.img */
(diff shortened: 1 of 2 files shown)


niftilib/nifti1_io.c
```diff
@@ -5211,7 +5206,7 @@ static int nifti_write_extensions(znzFile fp, nifti_image *nim)
 {
    nifti1_extension * list;
    char               extdr[4] = { 0, 0, 0, 0 };
-   int                c, size, ok = 1;
+   int                c;
```
hjmjohnson: Initializing `ok` is not needed here.  The value of the `ok` variable is unconditionally set at line 5238.

I was employing the "Prefer initialization to assignment" rule.  Here is the rationale that I subscribe to (and that is ubiquitous in all modern software development textbooks.), but I respect that you probably have different sensibilities as a seasoned C programmer:

=========

Variables should be declared as locally as possible.

Declaring variables "at the top of the function" is generally considered bad practice. Even in C89/90 language, where variables can only be declared at the beginning of the block, it is better to declare them as locally as possible, i.e. at the beginning of the smallest local block that covers the desired lifetime of the variable. Sometimes it might even make sense to introduce a "redundant" local block with the only purpose of "localizing" the variable declaration.

In C++ and C99, where it is possible to declare a variable anywhere in the code, the answer is pretty straightforward: again, declare each variable as locally as possible, and as close as possible to the point where you use it the very first time. The primary rationale for that is that in most cases this will allow you to supply a meaningful initializer to the variable at the point of declaration (instead of declaring it without initializer or with a dummy initializer).

As for memory usage, in general, a typical implementation will immediately (as you enter the function) allocate the maximum space required for all variables that exist at the same time. However, your declaration habits might affect the exact size of that space. For example, in this code

```c
void foo() {
  int a, b, c;

  if (...) {
  }

  if (...) {
  }
}
```
all three variables exist at the same time and generally the space for all three has to be allocated. But in this code

```c
void foo() {
  int a;

  if (...) {
    int b;
  }

  if (...) {
    int c;
  }
}
```
only two variables exist at any given moment, meaning that space for only two variables will be allocated by a typical implementation (b and c will share the same space). This is another reason to declare variables as locally as possible.

As with most recommendations, there follows an unstated: "unless you have a good reason to do otherwise in your situation". Efficiency might be enough of a reason to move variables outside of a loop in some cases. (Though for primitive types declaring them inside the loop won't make much of a difference.) And you could still surround the variables and loop with another scope to keep them from leaking into anywhere else, thus following the intent of the original idea.

hjmjohnson:
@afni-rickr Closing this PR without merging or rebasing.  A satisfactory alternative was presented in #88   

I added an explanation of my rationale for the other changes made.  I'm perfectly happy with ignoring this.

niftilib/nifti1_io.c
afni-rickr: I did understand this was more of a C++ thing.  In C, my preference strongly depends on the context, particularly on how long a function is, with the default of having variables in one place.  Only on occasion do I personally apply local defs, as it often suggests the current function is "too long", and might be better broken up.

PR #91, opened 2019-10-08
direct misc_tests command

afni-rickr:
Here is one command, duplicated from the d21_d script.  Does this change the coverage?

### DIFF
--- a/nifti2/CMakeLists.txt
+++ b/nifti2/CMakeLists.txt
@@ -133,7 +133,9 @@ if(NIFTI_BUILD_TESTING AND NIFTI_BUILD_APPLICATIONS)
           ${TEST_PREFIX}_tool_disp_ci ${TEST_PREFIX}_tool_disp_ts ${TEST_PREFIX}_tool_strip_extras
           PROPERTIES RESOURCE_LOCK Serial_zn1
           )
+  #==END NIFTI1 and NIFTI2 common tests ============================================
 
+  add_test( NAME ${TEST_PREFIX}_misc_tests COMMAND $<TARGET_FILE:${TOOL_NAME}> -run_misc_tests -debug 2 -infiles ${fetch_testing_data_SOURCE_DIR}/nifti_regress_data/e4.60005.nii.gz )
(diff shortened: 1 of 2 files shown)


hjmjohnson:
YES! This increases code coverage.

https://my.cdash.org/index.php?project=nifti_clib#!#Coverage

We are up to 48%!  Granted coverage statistics are only as good as the tests themselves.  Many of our test only confirms that "program ran without issuing a failure", and not that they actually did the correct action.

afni-rickr:
I think this confirms that the scripts do not get processed for coverage.  If we want them to, the individual commands must be pulled out as many separate tests.  That is unfortunate.  Again, even just running the four new c21 scripts gives me 90% functional coverage.  But I had to set something up to handle scripts.  I can copy some of the commands over.  Maybe we should add a couple of derivative datasets for the comparisons.

hjmjohnson:
Most of what is not covered is in the nifticdf directory:

https://my.cdash.org/viewCoverage.php?buildid=1712542&status=6&dir=nifticdf

The report for ./nifti2/nifti2_io.c seems perfectly reasonable for its coverage stat of 70% as having respected the shell scripts.   The 30% of uncovered code is all "debugging" and "failure" diagnostics.

https://my.cdash.org/viewCoverageFile.php?buildid=1712542&fileid=3631939

I curious if our reporting tools are using different metrics for computing coverage.

afni-rickr:
Oh, that does look good.  But then, did that c21_d command actually change anything?  That was identically pulled from a script.  Perhaps there is 70% coverage without the scripts.
The coverage does not seem to quite match the green lines.  Is -O0 set for the compilation, so that no optimization is done?
This looks great, thanks!

afni-rickr:
Oh, sorry, I am behind on the messages!  I guess you have found how to turn the scripts on, great!

afni-rickr:
By the way, my reporting tool is just aliasing nifti_tool to run lcov afterwards, and then accumulating the results and running genhtml.  That's an easy way to process entire scripts, and it seems to work very well, ending with a nice web page.

PR #109, opened 2020-06-15
Set so version in version header

leej3:
@afni-rickr, how does something like this look to you?


afni-rickr:
@leej3 These versions numbers seem good, but the code should be different.  It is preferable not to define variables in the .h files, and whatever is put there should have a unique name in any case.  Consider for nifti2_io_version.h :

```
#define NIFTI2_IO_VERSION_MAJOR 2
#define NIFTI2_IO_VERSION_MINOR 1
#define NIFTI2_IO_VERSION_MICRO 0
#define NIFTI2_IO_VERSION "2.1.0"
```
Or possibly _VERSION_STRING.  Whichever.  How does that seem?

leej3:
Thanks. That's better.

 I was going to leave out the major, minor,micro though. A little string parsing can extract those. And a clash between to two is possible.

afni-rickr:
Sure, either way is okay.  It is common for libraries to define MAJOR and MINOR versions separately.  But this seems good.

leej3:
@hjmjohnson, assuming this looks good to you, that's all the changes I had thought to make before the next release.

nifti2/nifti2_io_version.h
```diff
@@ -0,0 +1 @@
+#define NIFTI2_IO_VERSION "2.1.0"
```
hjmjohnson, line 1: I definitely prefer to have separate variables, and then a macro for the string version

Additionally, a note needs to be added to all the places that need to be kept manually in sync so that future developers change all the right code.

```c
/* NOTE:  When version values change, 4 places in the code need to  be kept in sync
 CMakeLists.txt  nifti2_io_version.h nifti1_io_version.h nifticdf_version.h
*/
#define NIFTI2_VERSION_TO_STRING(x) ITK_VERSION_TO_STRING0(x)
#define NIFTI2_VERSION_TO_STRING0(x) #x
#define NIFTI2_VERSION                                                                                                    \
  NIFTI2_VERSION_TO_STRING(NIFTI2_VERSION_MAJOR)                                                                             \
  "." NIFTI2_VERSION_TO_STRING(NIFTI2_VERSION_MINOR) "." NIFTI2_VERSION_TO_STRING(NIFTI2_VERSION_PATCH)
#define NIFTI2_SOURCE_VERSION "NIFTI2 version " NIFTI2_VERSION
```

afni-rickr:
Hi @hjmjohnson .  Yes, this all seems preferable to me, except for 2 bits.  I would rather have a prefix of NIFTI2_IO (and NIFTICDF, maybe), rather than just NIFTI, for the IO library version.  Because they are separate from the actual NIFTI versions.
But I am confused as to what might need to be manually kept in sync.  Only the macro values should change after this, independently per library directory.  Are you picturing what would happen if the actual macro names were changed?

leej3:
I've pushed what I understood by the suggestions. Let me know what if this is what you both think and I'll tidy up the changes and add comments as appropriate.

hjmjohnson:
> Hi @hjmjohnson . Yes, this all seems preferable to me, except for 2 bits. I would rather have a prefix of NIFTI2_IO (and NIFTICDF, maybe), rather than just NIFTI, for the IO library version. Because they are separate from the actual NIFTI versions.
> But I am confused as to what might need to be manually kept in sync. Only the macro values should change after this, independently per library directory. Are you picturing what would happen if the actual macro names were changed?

My concern is that he shared library versioning configured through the "project( VERSION " in cmake that is automatically captured from git into GIT_REPO_VERSION may become out of sync.

Personally, I think that we should not interrogate git for the version number, but rather I think we should manually have notes to indicate that these versions need to be updated when necessary.

leej3:
The GIT_REPO_VERSION approach was just an attempt to prevent disagreement between github releases and the cmake build. I'll change that back  (set the project version manually).

 With that said the SO versions will not necessarily be the same right? Perhaps I am misunderstanding something. For example, we currently have no breakage in API/ABI for any of the libraries except znzlib. In accordance with semantic versioning only znzlib would have a major version bump. I'm not sure if anyone has strong on opinions about what the release would be called but I thought 2.1.0 would be fine as the main library consumed by downstreams (libnifti2, or possibly libniftiio) would drive the overall package version. I don't have a sense of what is the most intuitive though. Happy to defer.

hjmjohnson:
@leej3 I think that this is an excellent step forward.  We should not try to solve all potential future problems now.  Let's push this through, and then revisit if needed in a few years from now.

leej3:
Great. Thanks

PR #111, opened 2020-06-18
Use ints for base version macros.  Add formatting.

afni-rickr:
@leej3 @hjmjohnson I believe this is what we wanted for the base macros, as ints (with added formatting).


leej3:
Sure. I can tweak the cmake build for that. Is this change to enforce  version numbers to integers rather than arbitrary strings?

afni-rickr:
Yes, they were were originally given as ints with a later conversion to strings in the TO_STRING macros.  That allows C code to use direct comparisons, if desired.

leej3:
Sounds good. Sorry for missing that detail!

afni-rickr:
Great, thanks!

PR #112, opened 2020-06-18
display new lib version string

afni-rickr:
... and apply the new version strings, seen via -nifti_ver in nifti_tool and nifti1_tool.


afni-rickr:
thanks, @hjmjohnson

PR #113, opened 2020-06-22
add manpage generation functionality

leej3:


leej3:
@hjmjohnson I've worked my way through packaging with conda, debian, and home-brew. The man generation is a little redundant in each instance. Is this a reasonable addition to reduce the burden elsewhere?

leej3:
@hjmjohnson I can't make sense of this error.

On OSX tests are failing because the nifti_tool can't be found:
https://travis-ci.com/github/NIFTI-Imaging/nifti_clib/jobs/352840211

Though the build shows no indication of this (although does still have an error):
https://my.cdash.org/buildSummary.php?buildid=1844387

Since its OSX I can't just download the environment. I have homebrew's cmake version 3.17 on OSX (instead of travis' 3.15.5) but can't reproduce these errors. 

Any ideas?

leej3:
Ok all solved now. I didn't use find_program for help2man so I struggled to interpret the error from the missing help2man dependency. I didn't realize that the help2man installation was failing silently on OSX. I switched the travis osx version to one that has a fixed homebrew installation.
 
I don't know whether setting NIFTI_INSTALL_NO_DOCS should be defaulted to ON so that help2man is only a build dependency if someone thinks to install man pages.

Also there is a warning that is not triggered on master that I do not understand: https://my.cdash.org/viewBuildError.php?type=1&buildid=1844859

PR #116, opened 2020-07-07
ENH: Prefer modern cmake "install COMPONENT" mechanism

hjmjohnson:
The modern cmake provides named COMPONENT for granular
control over which resources are installed.

https://cmake.org/cmake/help/v3.10/command/install.html

Remove outdated conditional variables that duplicate the modern
behavior.


hjmjohnson:
@leej3 We should probably consider removing "GIFTI_INSTALL_NO_DEVELOPMENT" from GIFIT as well now that COMPONENTS are used.

PR #117, opened 2020-07-08
add NIFTI_ECODE_MRS

afni-rickr:
ISMRM SG Code and Data sharing group are currently trying to establish an MR spectroscopy NIFTI data standard


afni-rickr:
also add a pointer cast

PR #119, opened 2020-07-20
fixes to install config for downstream usage

leej3:


leej3:
@hjmjohnson looks like there was a subtle bug in the Cmake files installed for downstreams. Unpleasant to debug. I used the following:

```
cd nifti_clib
mkdir build
mkdir downstream_build

# Run the following to test changes to nifti_clib installation config
cd build
ninja install 
ninja uninstall
rm -rf *
cmake -DDOWNLOAD_TEST_DATA=OFF -DBUILD_SHARED_LIBS=ON ..
ninja
cmake -DCOMPONENT=RuntimeBinaries -P cmake_install.cmake install
cmake -DCOMPONENT=Development -P cmake_install.cmake install
cmake -DCOMPONENT=RuntimeLibraries -P cmake_install.cmake install
cd ../parent_project_demo
rm -rf *
cmake -DFETCHCONTENT_SOURCE_DIR_NIFTI_CLIB=..  ../real_easy/parent_project_demo/
ninja
```

Homebrew won't accept a new formula with a patch and this is a fairly essential change. So I guess that means release 3.0.1?

leej3:
I was thinking of a test for this behavior too. Currently I have a test to confirm that installed binaries are correctly linking against the libraries. Perhaps this should be expanded to check that a minimal downstream project can link against the projects targets. I'm thinking something along the lines of:

```
cmake -DCMAKE_MODULE_PATH=/path/to/installed/config-files /path/to/minimal_project_demo
make
```

leej3:
Test added now.

PR #120, opened 2020-07-20
More minimal example of downstream usage

leej3:
I can put this in its own directory instead of overwriting the parent_project 
example if desired


leej3:
superseded by #119, the changes do not quite make sense separately

PR #125, opened 2020-08-26
Inconsistent definition of NIfTI-1 image struct

jonclayden:
In the NIfTI-1 library, and specifically in `nifti1_io.h`, the `nifti_image` struct has an `nvox` member of type `size_t`. By contrast, in `nifti2_io.h`, the definition of the `nifti1_image` struct switches the type to `int64_t`. However, these two types are not equivalent. `size_t` is always an _unsigned_ type – although it is unlikely that the sign bit would become relevant for any real-world image. More problematic in practice is that `sizeof(size_t)` is 4 (rather than 8) on some platforms, so the redefinition will change the width of the type, leading to an unnecessary binary incompatibility with objects created using the NIfTI-1 library. This leads to a segfault in a wrapper library of mine that provides access to both versions of the NIfTI library.

This fairly trivial PR therefore changes the definition back, to use `size_t`.

### DIFF
--- a/nifti2/nifti2_io.h
+++ b/nifti2/nifti2_io.h
@@ -227,7 +227,7 @@ typedef struct {
   int nv ;                      /*!< dimensions of grid array             */
   int nw ;                      /*!< dimensions of grid array             */
   int dim[8] ;                  /*!< dim[0]=ndim, dim[1]=nx, etc.         */
-  int64_t nvox ;                /*!< number of voxels = nx*ny*nz*...*nw   */
+  size_t nvox ;                 /*!< number of voxels = nx*ny*nz*...*nw   */
   int nbyper ;                  /*!< bytes per voxel, matches datatype    */
   int datatype ;                /*!< type of data in voxels: DT_* code    */
 


afni-rickr:
Hi @jonclayden, nvox was changed to not limit NIFTI datasets to 4 billion voxels.  That was considered to be an improvement, though changing it under niftilib would have introduced a lack of backwards compatibility in the nifti-1 library (which wasn't required under nifti-2).
It is true that there may be no particularly big reason to have the nifti1_image struct be possibly different under NIFTI-2.  Similarly, there wasn't considered any big reason to have them be identical.
But the main reason for the change is so the library could correctly store a large dataset in a nifti1_image structure.  It is not clear what is really better here.  @hjmjohnson, do you have any thoughts?

jonclayden:
I see – I had assumed (wrongly, clearly) that the idea was to standardise on fixed-width integer types for everything potentially bigger than `int`.

Ultimately I defer to your experience, but `nifti1_image` isn't used by any API functions and clients of the NIfTI-2 library explicitly choose `nifti2_image` as their standard data structure, so it's hard to imagine conditions in which they might use a slightly tweaked version of `nifti1_image`.

On the other side of the coin, the change makes it harder for NIfTI-2 clients to work with in-memory `nifti1_image` objects created by the NIfTI-1 library. I care about that use-case, so I will make this change in my redistribution of the library regardless, but I recognise that it's unusual.

afni-rickr:
Fixed-width integer sizes were certainly desired, and size_t was a mistake in that regard.  Types like int64_t were newer C than what we had originally tried to require for the NIFTI library, which confused these things.

hjmjohnson:
@jonclayden Can this be closed?  It appears that this turns out to not be relevant to merge into the primary codebase.

jonclayden:
I think this ends up being mostly a question of taste. It can create problems in practice, as I have found, but only in quite esoteric circumstances – so on that basis I'm happy to follow your lead and close this. Thanks for taking the time to feed back.

PR #126, opened 2020-08-27
N2 updates

afni-rickr:
Combine pull request 125 with corresponding updates to nifti_tool and test of MAKE_IM files.


afni-rickr:
@jonclayden Finally got back to this, sorry.  Please let me know if you have any comments, thanks.

jonclayden:
Thanks. Yes, this looks good to me – it makes sense for `nifti_tool` to always produce NIfTI-1 output when performing an operation that explicitly targets the NIfTI-1 data structure.

PR #140, opened 2022-02-02
Generate pkgconfig files

tintou:
This allow to easily include the libraries in other projects.

See https://www.freedesktop.org/wiki/Software/pkg-config/ for more information
about what pkg-config is.


tintou:
@kleisauke thanks! this makes sense, I've added your patch

tintou:
Rebased to latest master and ready to review

PR #142, opened 2022-02-24
ENH: nifti_tool -copy image

afni-rickr:
Add -copy_image option and the ability to convert datatype to nifti_tool.
We could consider moving the conversion function to the library.  That would affect the library version.


afni-rickr:
I believe this just needs the actual data conversion to be done.  That could be done in pieces, since there are D(D-1) cases to consider (D=num valid datatypes).  They do not all need to allowed, even if possible.

afni-rickr:
This seems basically ready.

hjmjohnson:
@afni-rickr Are tests added to verify that the changed behavior works?  Future maintenance will be easier if each feature is has an appropriate test are available for the changed command-line options.

afni-rickr:
@hjmjohnson I have been wondering how to do that effectively, so testing has been local so far.  The only reasonable tests that I can see are doing 2 conversions that invert, and seeing that the results are unchanged (across some types).  Otherwise, an external calculation program would be needed 
Currently I use AFNI tools for local tests, to include mod() and int() truncation.  For example, I test int16 to uint8 by comparing with mod(256), and float to int by comparing with truncation.  But it helps to have external tools for that.
I don't test all type pairs, since they use the exact same macro for the conversion.
Can you see other tests that would not need an external calculation?
Without that, an inversion test could be applied for some types in the cases of general nim and NBL reading (e.g. -copy_image/-cbl).  Can you think of anything better?  Thanks!

hjmjohnson:
@afni-rickr I was hoping for at least an exercising of the command line arguments in the test.  

If the generated files are expected to be bitwise equivalent on all platforms, a hash could be computed for a known good result.

https://cmake.org/cmake/help/v3.8/command/file.html

I find regression tests helpful for ensuring that future changes to the codebase do not change the behavior.  Of course, your external tools are still necessary to ensure the current behavior is correct.

afni-rickr:
@hjmjohnson A bitwise test (even using diff or cmp) works if the endian matches.  To do that, I could copy input to a new int16 one (to apply current endian), go to int64 and float32 and back (to int16) and 'cmp' the results.  There should be no binary differences.  I will put in something like that based on my current regression tests.

hjmjohnson:
@afni-rickr I think a little endian only test is sufficient.
```cmake
include (TestBigEndian)
TEST_BIG_ENDIAN(IS_BIG_ENDIAN)
if(IS_BIG_ENDIAN)
 message(STATUS "BIG_ENDIAN")
else()
 message(STATUS "LITTLE_ENDIAN")
endif()
```

Then a copy with a hash test would also be sufficient. 

I am fine with your proposal as well.  

Thank you for adding a test.

afni-rickr:
@hjmjohnson It looks like there is a mistake at the end of install_linking_test.sh, not referencing back into the nifti_clib directory with the final cmake command.  This is confusing as I was not aware there is a lingering issue.  I will check this file in.
Okay, does "fix path to downstream dir" seem appropriate?  That looks like it was needed in master.  Thanks.

afni-rickr:
@hjmjohnson Okay, I have added a test script (to be clear, I had planned to do so, but wanted your thoughts on what seemed reasonable once the code was ready).  This part seems good now, too.  Thanks!

hjmjohnson (approved):
Thank you for adding the test.  I see no problems with the changes.

afni-rickr:
Great, thanks!

PR #143, opened 2022-03-01
use copy_image

afni-rickr:
Update Updates.txt, and replace -copy_im with -copy_image in test c15.
Note -copy_im works as -cbl, which collapses upper dims (so it isn't necessarily a copy).


afni-rickr:
This should hopefully be the final update for the copy/convert ENH.

hjmjohnson (approved):

PR #146, opened 2022-05-04
Nifti error

hjmjohnson:
When writing fails, there needs to be a mechanism for detecting the errors.

### DIFF
--- a/niftilib/nifti1_io.c
+++ b/niftilib/nifti1_io.c
@@ -2170,7 +2170,7 @@ void nifti_swap_4bytes( size_t n , void *ar )    /* 4 bytes at a time */
        tval = *cp1;  *cp1 = *cp2;  *cp2 = tval;
        cp0 += 4;
    }
-   }
+}
 
 /*----------------------------------------------------------------------*/
 /*! swap 8 bytes at a time from the given list of n sets of 8 bytes
@@ -2192,7 +2192,7 @@ void nifti_swap_8bytes( size_t n , void *ar )    /* 8 bytes at a time */
        }
        cp0 += 8;
    }
-   }
+}
 
 /*----------------------------------------------------------------------*/
 /*! swap 16 bytes at a time from the given list of n sets of 16 bytes
@@ -2212,7 +2212,7 @@ void nifti_swap_16bytes( size_t n , void *ar )    /* 16 bytes at a time */
        }
        cp0 += 16;
    }
-   }
+}
 
 #if 0  /* not important: save for version update     6 Jul 2010 [rickr] */
 
@@ -2254,7 +2254,7 @@ void nifti_swap_Nbytes( size_t n , int siz , void *ar )  /* subsuming case */
         fprintf(stderr,"** NIfTI: cannot swap in %d byte blocks\n", siz);
         break ;
    }
-   }
+}
 
 
 /*-------------------------------------------------------------------------*/
@@ -2410,7 +2410,7 @@ void old_swap_nifti_header( struct nifti_1_header *h , int is_nifti )
      nifti_swap_4bytes(4,h->srow_y);
      nifti_swap_4bytes(4,h->srow_z);
    }
-   }
+}
 
 
 #define USE_STAT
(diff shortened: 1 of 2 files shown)


hjmjohnson:
@afni-rickr This proposed change assists with resolving an issue in ITK where we can not programmatically detect when an error occurs during writing.

afni-rickr:
@hjmjohnson Indeed, this seems good.  Rarely do void functions seem appropriate, and surely not here.
Should I made some corresponding changes in this particular branch, checking the result?

hjmjohnson:
@afni-rickr I do not have a good suggestion.  Somehow we need to have a programmatic way of detecting when an error occurs.  Printing a message is insufficient.

I am concerned that the global variable for managing error states will cause many problems in a multi-threaded environment.

=====
To maintain backward compatibility, we could keep the original functions as they existed previously, and provide new functions (perhaps with limited capabilities) that are capable of reporting error states.  Perhaps we would want to have compile-time cmake options for deprecating the original funcitons.

afni-rickr:
@hjmjohnson Indeed, then maybe we should just create a new function, perhaps nifti_image_write_hdr_img_status(), that could work the way we think it ought to, starting with just small changes to nifti_image_write_hdr_img2(), but with a new function name.  Then both nifti_image_write() and nifti_image_write_hdr_img2() could just be tiny wrappers to call the _status() version.
Then nifti_image_write() would be able to return the status from the modified function.
The general effect of that on the library would just be to
1. as you did, have nifti_image_write() return status
2. add a new function nifti_image_write_hdr_img_status()
That would maintain backward compatibility with the library, except that some applications would start getting warnings for throwing away the return value from nifti_image_write().
If we further cheated and made the _status() function static (which we could have done with many of these functions), we would not even need to claim a new function.
If we add a new function, that might imply a major version update.  I forget.
Does that seem like a reasonable plan?  I would be willing to fire it off, maybe in a fresh branch, just for clarity.

afni-rickr:
@hjmjohnson I have pondered this a bit more, and am worried about changing the return type of the function.  Wouldn't changing from void to int affect the call stack and possibly break things?  That is, if a program were compiled against a library with a void function, and then it starts returning an int in the newer version, would that not create a possible runtime error, depending on the architecture?

dzenanz:
Wouldn't compiler optimize away an unused return value?

afni-rickr:
@dzenanz Just to be clear, my concern is for an already existing application that was compiled against the void version now being run against the int version (i.e. the library gets upgraded, but the application does not get recompiled).  The library would not expect the return value to ignored, and the app would not expect the library function to return anything.  Or does this simply require that any app must be recompiled against the newer version of the library, which would presumably imply a major version update of the library, even for such a tiny change?

afni-rickr:
Thanks @dzenanz. Out of curiosity, under what case would a linux .so library fall?  But indeed, so this would be a bigger change than adding a new function.
@hjmjohnson Given these thoughts, do you prefer a new function or a changed one?  While the changed return type seems nicer, it is more dangerous from a backwards-compatibilty standpoint.  Either way, this will require some sort of version update, which means we should probably make some of the other changes, too (like what I promised under #138).

dzenanz:
DLL is a Windows equivalent of Linux `.so`. I think crashing during app load and working correctly are the most likely options for Windows. A single return value is usually returned in a register (EAX), so it would mostly get implicitly ignored by the unsuspecting application.

afni-rickr:
Thanks.  I guess use of a register for the return value (vs stack) is the basic question regarding whether it would work, which is presumably a system and optimizer-dependent compiler choice.
So maybe I will actually proceed with the new function route, and we can ponder where to go from there.

hjmjohnson:
@afni-rickr A completely new function would be just fine by me.  We just need *an* interface to exists that we wrap to accomplish the reading of nifti files with error checking.

The nifti api is not exposed in ITK.

hjmjohnson:
@afni-rickr Could you please make the new function with an appropriate naming scheme so that I may update the outstanding issue in ITK?

afni-rickr:
@hjmjohnson Sorry for being slow with this, I will attack it soon.  Do you have an opinion for the new top-level function name, such as nifti_image_write_stat (or status)?

PR #147, opened 2022-06-06
ENH: add nifti_image_write_status

afni-rickr:
add NIFTI-1 and -2 versions of nifti_image_write_status

### DIFF
--- a/nifti2/nifti_tool.c
+++ b/nifti2/nifti_tool.c
@@ -7123,14 +7123,16 @@ int act_cbl( nt_opts * opts )
 
    /* and finally, write out results */
    if( err == 0 && nifti_nim_is_valid(nim, g_debug) )
-      nifti_image_write_bricks(nim, &NBL);
+      if( nifti_image_write_bricks_status(nim, &NBL) ) {
+         err++;
+      }
 
    nifti_image_free(nim);
    nifti_free_NBL(&NBL);
    free(fname);
    free(selstr);
 
-   return 0;
+   return err;
 }
 
 
--- a/nifti2/nifti_tool.c
+++ b/nifti2/nifti_tool.c
@@ -2291,7 +2291,13 @@ int act_add_exts( nt_opts * opts )
          fprintf(stderr,"+d writing %s with %d new extension(s)\n",
                  opts->infiles.list[fc], opts->elist.len);
 
-      nifti_image_write(nim);
+      if( nifti_image_write_status(nim) ) {
+         fprintf(stderr,"** failed to write image %s\n",
+                 opts->infiles.list[fc]);
+         nifti_image_free(nim);
+         return 1;
+      }
+
       nifti_image_free(nim);
    }
 
@@ -2403,7 +2409,11 @@ int act_strip( nt_opts * opts )
          fprintf(stderr,"+d writing %s without extensions or 'descrip'\n",
                  nim->fname);
 
-      nifti_image_write(nim);
+      if( nifti_image_write_status(nim) ) {
+         fprintf(stderr,"** failed to write image %s\n", nim->fname);
+         nifti_image_free(nim);
+         return 1;
+      }
 
       if( g_debug > 3 ) nifti_image_infodump(nim);
       nifti_image_free(nim);
@@ -2477,7 +2487,12 @@ int act_rm_ext( nt_opts * opts )
          fprintf(stderr,"+d writing %s with %d fewer extension(s)\n",
                  nim->fname, ext_ind == -1 ? num_ext : opts->elist.len);
 
-      nifti_image_write(nim);
+      if( nifti_image_write_status(nim) ) {
+         fprintf(stderr,"** failed to write image %s\n", nim->fname);
+         nifti_image_free(nim);
+         return 1;
+      }
+
       nifti_image_free(nim);
    }
 
@@ -3337,7 +3352,13 @@ int act_mod_hdrs( nt_opts * opts )
          }
          dupname = nifti_strdup(nim->fname);  /* so we know to free it */
          fname = dupname;
-         nifti_image_write(nim);  /* create the duplicate file */
+         /* create the duplicate file */
+         if( nifti_image_write_status(nim) ) {
+            fprintf(stderr,"** failed to write image %s\n", nim->fname);
+            nifti_image_free(nim);
+            return 1;
+         }
+
          /* if we added a history note, get the new offset into the header */
          /* mod: if the new offset is valid, use it    31 Jan 2006 [rickr] */
          if( nim->iname_offset >= 348 ) nhdr->vox_offset = nim->iname_offset;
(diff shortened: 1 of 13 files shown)


gdevenyi:
Just a thought, have you considered converting `nifti_image_write` to a thin wrapper which calls `nifti_image_write_status` and discards the return bits? This could reduce code duplication and ensure a unified path for code testing etc.

afni-rickr:
@gdevenyi Yes, there are many such bits to possibly do.  Currently, nifti_image_write does call _status, except that it goes through 2 intermediate layers first.  We could cut out those layers, though it is weighed against modifying functions that work as they are.  But if you think _write should call _status, I will make that change.  I will duplicate most of this in the NIFTI-1 library, too.  Thanks.

gdevenyi:
I think for consistency/uniformity it makes sense for `nifti_image_write` and `nifti_image_write_status` to take the same pathways through the code wherever possible.

My only concern is that based on your description, changing `nifti_image_write` to wrap `nifti_image_write_status` is a big-ish change to the existing codeflow, as they're implemented differently? I can see either side of this discussion winning out, perhaps it depends on what the next version # of the release would be :)

afni-rickr:
The addition of _engine is already modifying the code for most of the image_write functions, including nifti_image_write, so this is already a somewhat impactful change.  Since nifti_image_write is already calling the engine (indirectly), it would make sense to just be direct about it.
I am thinking to make a similar nifti_image_write_bricks_status function, too, esp. since write_bricks is already relying on _engine.
So the _engine function is the real change (for NIFTI-1 and -2), but it will lead to lots of little changes, including in the examples.

nifti2/nifti_tool.c
```diff
@@ -2291,7 +2291,13 @@ int act_add_exts( nt_opts * opts )
          fprintf(stderr,"+d writing %s with %d new extension(s)\n",
                  opts->infiles.list[fc], opts->elist.len);
 
-      nifti_image_write(nim);
+      if( nifti_image_write_status(nim) ) {
+         fprintf(stderr,"** failed to write image %s\n",
```
hjmjohnson, line 2295: Should this be protected by "if( g_opts.debug > 1 )"?

nifti2/nifti_tool.c
```diff
@@ -2403,7 +2409,11 @@ int act_strip( nt_opts * opts )
          fprintf(stderr,"+d writing %s without extensions or 'descrip'\n",
                  nim->fname);
 
-      nifti_image_write(nim);
+      if( nifti_image_write_status(nim) ) {
+         fprintf(stderr,"** failed to write image %s\n", nim->fname);
```
hjmjohnson, line 2413: Should this be protected by "if( g_opts.debug > 1 )"?

nifti2/nifti_tool.c
```diff
@@ -2477,7 +2487,12 @@ int act_rm_ext( nt_opts * opts )
          fprintf(stderr,"+d writing %s with %d fewer extension(s)\n",
                  nim->fname, ext_ind == -1 ? num_ext : opts->elist.len);
 
-      nifti_image_write(nim);
+      if( nifti_image_write_status(nim) ) {
+         fprintf(stderr,"** failed to write image %s\n", nim->fname);
```
hjmjohnson, line 2491: Should this be protected by "if( g_opts.debug > 1 )"?

nifti2/nifti_tool.c
```diff
@@ -3588,7 +3615,13 @@ int act_swap_hdrs( nt_opts * opts )
          }
          dupname = nifti_strdup(nim->fname);  /* so we know to free it */
          fname = dupname;
-         nifti_image_write(nim);  /* create the duplicate file */
+         /* create the duplicate file */
+         if( nifti_image_write_status(nim) ) {
+            fprintf(stderr,"** failed to write image %s\n", nim->fname);
```
hjmjohnson, line 3620: Should this be protected by "if( g_opts.debug > 1 )"?

nifti2/nifti_tool.c
```diff
@@ -7134,7 +7173,13 @@ int act_copy( nt_opts * opts )
       }
 
       /* and write out results */
-      if( nifti_nim_is_valid(nim, g_debug) ) nifti_image_write(nim);
+      if( nifti_nim_is_valid(nim, g_debug) ) {
+         if( nifti_image_write_status(nim) ) {
+            fprintf(stderr,"** failed to write image %s\n", nim->fname);
```
hjmjohnson, line 7178: Should this be protected by "if( g_opts.debug > 1 )"?

nifti2/nifti_tool.c
```diff
@@ -7194,7 +7239,13 @@ int act_cci( nt_opts * opts )
    if(g_debug>2) disp_field("new nim:\n",g_nim2_fields,nim,NT_NIM_NUM_FIELDS,1);
 
    /* and finally, write out results */
-   if( nifti_nim_is_valid(nim, g_debug) ) nifti_image_write(nim);
+   if( nifti_nim_is_valid(nim, g_debug) ) {
+      if( nifti_image_write_status(nim) ) {
+         fprintf(stderr,"** failed to write image %s\n", nim->fname);
```
hjmjohnson, line 7244: Should this be protected by "if( g_opts.debug > 1 )"?

niftilib/nifti1_io.c
```diff
@@ -2254,7 +2256,7 @@ void nifti_swap_Nbytes( size_t n , int siz , void *ar )  /* subsuming case */
         fprintf(stderr,"** NIfTI: cannot swap in %d byte blocks\n", siz);
```
hjmjohnson, line 2256: Should this be protected by "if( g_opts.debug > 1 )"?

hjmjohnson (changes requested):
Overall looks good.  I am concerned about printing to stderr unconditionally in some cases.

afni-rickr:
For the most part, serious errors have always been printed to stderr unconditionally.
There are some cases of `if( g_opts.debug > 0 )`, but not so many.  Also, the _engine, as with the old hdr_img2 code will generally print the reason for failure.  Those could be wrapped in a test for debug > 0 (probably not 1), but that would be a change.

hjmjohnson:
@afni-rickr I am OK with not changing historical behavior in the PR.

I recommend that a future PR consider not writing to terminal stderr from library functions.  Where should those messages be displayed for a GUI application?

afni-rickr:
@hjmjohnson sure, a future PR preventing default stderr writing from the library seems reasonable.  For a GUI, are those messages simply not seen?  If that is what happens, such a PR would not affect them.  We can ponder it.

afni-rickr:
@hjmjohnson I am going to make a quick parallel update for nifti_image_write_bricks (which is already calling _engine, eventually).  Would you like a test added for such a failure case?  I have done local testing using a read-only output directory.

hjmjohnson:
@afni-rickr I think that test may be hard to write.  I'll be satisfied with historical testing that did not test the error status.

Thank you for you assistance with making these changes.

afni-rickr:
@hjmjohnson Okay, I think this is ready.  Would you like a last peek at the recent commits?  Please feel free to merge if you are happy (or at least moderately satisfied :).

hjmjohnson (approved):

PR #152, opened 2022-08-02
ENH: Hide advanced cmake settings

hjmjohnson:
The NIFTI_SYSTEM_MATH_LIB variable does not normally need
to be presented to the user as a standard configuration
option.


hjmjohnson:
@dzenanz @afni-rickr

hjmjohnson:
Hiding this is important to avoid exposing by default to third-party integrators.

afni-rickr (approved):

dzenanz (approved):

PR #156, opened 2022-08-12
WIP: More warning fixes

seanm:


seanm:
@hjmjohnson @afni-rickr some of these may be more controversial... as they change API or behaviour slightly.

Thoughts?

niftilib/nifti_tester001.c
```diff
@@ -117,7 +117,7 @@ nifti_image * generate_reference_image( const char * write_image_filename , int
 }
 
 
-void compare_reference_image_values(nifti_image const * const reference_image, nifti_image const * const reloaded_image, int * const Errors)
+static void compare_reference_image_values(nifti_image const * const reference_image, nifti_image const * const reloaded_image, int * const Errors)
```
hjmjohnson, line 120: These are VERY NICE!  Keep the API clean, and avoid global namespace pollution.

niftilib/nifti1_io.c
```diff
@@ -2441,7 +2441,7 @@ int nifti_get_filesize( const char *pathname )
 
    if( pathname == NULL || *pathname == '\0' ) return -1 ;
    ii = stat( pathname , &buf ); if( ii != 0 ) return -1 ;
-   return (unsigned int)buf.st_size ;
+   return (int)buf.st_size ;
```
hjmjohnson, line 2444: Nice catch!

nifti2/nifti2_io.c
```diff
@@ -3540,7 +3540,7 @@ char * nifti_find_file_extension( const char * name )
                  ext);
          return NULL;
       }
-      else return (char *)ext; /* Cast away the constness of the input parameter */
+      else return ext;
```
hjmjohnson, line 3543: So much better!

nifti2/nifti2_io.c
```diff
@@ -564,7 +564,7 @@ void nifti_disp_lib_hist( int ver )
 *//*--------------------------------------------------------------------*/
 void nifti_disp_lib_version( void )
 {
-   printf("%s, compiled %s\n", gni_version, __DATE__);
```
hjmjohnson, line 566: If date is useful for some reason here, it should be generated and passed in from a cmake variable that can be set and fixed.

hjmjohnson (approved):
@seanm These are very nice changes from a software engineering point of view.

I'll let @afni-rickr comment on the historical context.

nifti2/nifti2_io.c
seanm: I don't think this will break many clients, and it's just so wrong it's worth fixing IMHO.  It's also the *only* -Wconst-qual warning left.

seanm:
@afni-rickr care to weigh in on this WIP?

hjmjohnson (approved):

afni-rickr:
Sorry for being slow.  I am not sure how much to go into right now, but these changes are not thrilling to me.
- static functions in main programs: not clear why
- addition of assert: do not really like library deciding to crash
- removal of __DATE__: why is this important? not everyone uses cmake
- functions returning const char *: feeding this const virus does not seem necessary
- signed/unsigned casting: I do not think those were mistakes
I will try to post more specific comments/questions...

seanm:
>static functions in main programs: not clear why

Superficially: to fix `-Wmissing-prototypes` warnings.  It prevents private symbols from being exported, thus allowing some complier/linker optimizations.

>addition of assert: do not really like library deciding to crash

There will be no crash in reality, because the `default` case of that `switch` statement is impossible.  The `assert(0)` just makes the compiler understand that it's impossible.

>removal of DATE: why is this important? not everyone uses cmake

cmake is irrelevant. Removal of `__DATE__` is for reproducible builds. Some motivation discussed here for example: https://reproducible-builds.org/

>functions returning const char *: feeding this const virus does not seem necessary

:) Yes, const can be like a virus...

But casting away const is a very big foot gun.  It might technically be safe in these instances, but as these are the *only* remaining occurrances of `-Wconst-qual` warnings, there's no big cascade of additional `const` to come.

Having a codebase build cleanly with `-Wconst-qual` I think is a must these days.

>signed/unsigned casting: I do not think those were mistakes

Could you elaborate?

nifti2/clib_02_nifti2.c
```diff
@@ -19,7 +19,7 @@
 #include <nifti2_io.h>
 int disp_float(nifti_image * nim, int vol, int slice, int line, int offset);
 
-int show_help( void )
+static int show_help( void )
```
afni-rickr, line 22: this is a main program, where static should not really matter

nifti2/nifti2_io.c
afni-rickr: making reproducible builds is a good reason to omit `__DATE__`, thanks

nifti2/nifti2_io.c
```diff
@@ -3398,7 +3399,7 @@ int64_t nifti_get_volsize(const nifti_image *nim)
 
     \return 1 on existence, 0 otherwise
 *//*--------------------------------------------------------------------*/
-int nifti_fileexists(const char* fname)
+static int nifti_fileexists(const char* fname)
```
afni-rickr, line 3402: I guess we can see if anyone complains, but this removes the function from the library.  An alternate change might be to add the prototype to the .h file.

seanm: So this illustrates why having no `-Wmissing-prototypes` warnings in the codebase is desirable. If this is was meant to be public API, we found a bug: namely that it was missing from the header file.

I can alter this change to put it in the header if you want.  I didn't know if it was meant to be public API or not.

nifti2/nifti2_io.c
afni-rickr: There is only one other exported "const char *" function in nifti2_io.c.  This will force calling code to deal with the viral const.  The strstr method (for example) seems nicer (`man strstr`), where the library allows one to use either const or not.  I do not really see this as wrong.

niftilib/nifti_tester001.c
```diff
@@ -7,7 +7,7 @@ enum NIFTITEST_BOOL {
   NIFTITEST_FALSE=0
 };
 
-void _PrintTest(const int line,const char * message,const int FailureOccured, const enum NIFTITEST_BOOL isFatal,int *ErrorAccum)
+static void _PrintTest(const int line,const char * message,const int FailureOccured, const enum NIFTITEST_BOOL isFatal,int *ErrorAccum)
```
afni-rickr, line 10: This is a main program, so static should not matter.

niftilib/nifti1_io.c
afni-rickr: I think this `unsigned` was intentional.  From what I recall, some compilers get mad about changing both the sign and the size of an integer, and this may have been done as a 2-step process.
Maybe a better cast would be: (int)(unsigned int)

niftilib/nifti1_io.c
```diff
@@ -3652,7 +3656,7 @@ nifti_image* nifti_convert_nhdr2nim(struct nifti_1_header nhdr,
       *   the qform_code will be zero, at which point you can check
       *   analyze75_orient if you care to.
       */
-     unsigned char c = *((char *)(&nhdr.qform_code));
+     unsigned char c = *((unsigned char *)(&nhdr.qform_code));
```
afni-rickr, line 3659: This is okay, but does not really matter.  The original field is signed in C, which is probably why the original cast was signed.  If any compilers care about that aspect, this could be better as:
`(unsigned char)*((char *)(&nhdr.qform_code))`
which might have been considered overkill.

nifti2/clib_02_nifti2.c
seanm: Indeed, there should be no change in behaviour here.

The motivation is to fix all `-Wmissing-prototypes` warnings because that warning in general can and does find real bugs.

It's just good practice for symbols not needed outside the translation unit to be declared `static`.

nifticdf/nifticdf.c
```diff
@@ -10199,7 +10199,7 @@ static pqpair student_s2pq( double xx , double dof )
 
 /*------------------------------*/
 
-double student_pq2s( pqpair pq , double dof )
+static double student_pq2s( pqpair pq , double dof )
```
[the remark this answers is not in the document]
seanm: That would also require moving `typedef struct { double p,q ; } pqpair ;` to the header...

niftilib/nifti_tester001.c
seanm: Indeed, there should be no change in behaviour here.

The motivation is to fix all `-Wmissing-prototypes` warnings because that warning in general can and does find real bugs.

It's just good practice for symbols not needed outside the translation unit to be declared `static`.

hjmjohnson: @seanm Thanks.  I agree with you.  

While I agree that it does not matter to the functioning of the program, it does make maintaining the package more difficult for developers who utilize advanced IDEs and code checkers. Adding static in places where it could be optional makes long-term maintenance easier for new (and some old) developers.

I feel similarly about "const". Using const makes refactoring and other long-term code maintenance concerns less dangerous.  It conveys to new developers a set of constraints that can make understanding the code easier.

afni-rickr:
Finally going through this in reasonable detail... it seems more clear to group comments here, rather than making either of you search through the list of conversations.

nifti2/clib_02_nifti2.c:    show_help()
niftilib/nifti_tester001.c: generate_reference_image()
Yes, of course, static is fine there. 
I was thrown by the comments on cleaning up the global namespace (and the translation unit), along with the library functions missing them, and did not even notice that these actually had no prototypes until finally going through this in more detail.  Adding actual prototypes at the top might be preferable, but either way is fine.

nifti2/nifti2_io.c: nifti_fileexists()
nifti2/nifti2_io.c: nifti_extension_size()
Removing symbols from the library could be far more strongly justified as a real bug by someone using them.  Removing them breaks backwards compatibility, and it works as it is.  Clearly they should have prototypes in any case.
For nifti_fileexists(), putting the prototype in nifti2_io.h seems appropriate.  That looks like a function people would use.  But it is hard to say about nifti_extension_size().  Maybe that function would be okay to remove from the library exports (though it should still get a (static) prototype at the top of nifti2_io.c).  Either making it static or putting it in the .h seems okay, though we will have to make a version comment if it is removed from the export list.

nifti2/nifti2_io.c: nifti_find_file_extension()
It would be nice to look into this further.  Maybe you mean -Wcast-qual here.  In any case, I would much rather resolve such warnings with internal casting than to inflict const on the calling packages, speaking as one who has gone through that.  Maybe this change should be left out and noted for further investigation.  Or we could take the time to figure it out now... time is hard to come by.

nifticdf/nifticdf.c: student_pq2s()
Yes indeed, that function looks intended to be local.  I would prefer to give it a prototype at the top of nifticdf.c (including static), but that file (from external sources) does not seem to do that for anything.  So indeed, just adding 'static' seems fine.

niftilib/nifti1_io.c: nifti_get_filesize()
To reduce similar confusion about whether the casts were mistakes in the future, this should probably be done explicitly, as something like:
```
/* alter size and sign in separate casts to prevent compiler whining */
return (int)(unsigned int)but.st_size;
```

niftilib/nifti1_io.c: nifti_convert_nhdr2nim()
Similar to nifti_get_filesize().
```
/* alter size and sign in separate casts to prevent compiler whining */
unsigned char c = (unsigned char)*((char *)(&nhdr.qform_code));
```
niftilib/nifti_tester001.c: _PrintTest()
Adding static or a prototype is great here, of course.
Also, that leading '_' may go against the C standard.  It looks like the '_' version was added to take the `__LINE__` parameter after the fact.  So how about this being done with a different name (we tend to add and 'engine' suffix, but whatever is okay):
`static void PrintTest_eng(...);`

Thanks.

seanm:
1.

>nifti2/nifti2_io.c: nifti_fileexists()
>nifti2/nifti2_io.c: nifti_extension_size()
>Removing symbols from the library could be far more strongly justified as a real bug by someone using them. Removing them breaks backwards compatibility, and it works as it is.

Is it not exceedingly improbable that anyone is using them, given that there's no declaration in the .h file?

2. I've dropped the -Wcast-qual commit. Will make a separate PR for further discussions.
3. I've dropped the casting changes.
4. I added the PrintTest_eng rename.

niftilib/nifti1_io.c
seanm: I've dropped this change.

niftilib/nifti1_io.c
seanm: I've dropped this change.

nifti2/nifti2_io.c
seanm: I'll move these const changes to a new PR for fresh arguments :)

afni-rickr:
> > nifti2/nifti2_io.c: nifti_fileexists()
> > nifti2/nifti2_io.c: nifti_extension_size()
> > Removing symbols from the library could be far more strongly justified as a real bug by someone using them. Removing them breaks backwards compatibility, and it works as it is.
> 
> Is it not exceedingly improbable that anyone is using them, given that there's no declaration in the .h file?

It would be very surprising if no one is using nifti_fileexists(), in particular.  Missing a prototype in the .h file is not going to deter someone who wants to use it.  I recall many instances of function prototypes being missing from standard (system installable) packages.  If it is the function you want to use and it is in the library, you use it.

hjmjohnson:
@afni-rickr If we want to support nifti_fileexists, then we need to put it in the public interface in the standard way.

Either make it static to declare that it is not available, or make it available in the .h file. 

Just because some "can be done" does not mean that we should allow or support it.

seanm:
>Either make it static to declare that it is not available, or make it available in the .h file.

Exactly. Just a question of deciding if it's supposed to be public API or not.  Sounds like @afni-rickr is saying it's meant to be public.

afni-rickr:
Removing symbols should be considered a moderately big deal, as it breaks backward compatibility.  For exported functions missing a prototype, the "default" should be to simply add the prototype, not to take them out of the library.  And I see no good reason to remove nifti_fileexists().
Even extension_size, it is really preferable to remove it?  I am not particularly against that, but anything removed at this point needs a clear entry perhaps in Updates.txt, at least.

seanm:
I mean, arguing this is not a hill I want to die on, but, in my view, anyone using a function not declared in a public header is asking for trouble and should not expect backwards compatibility when using private API.

I'll note that ITK and libMINC (the only niftilib users I know) are not using it.

Finally, that function should be discouraged anyway, as it's prone to [TOCTOU](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use) bugs.

afni-rickr:
I would rather not punish someone for using a function that exists in the library.  It is odd that the function tries to actually open the file, rather than simply call stat(), say.  But it does not really seem to be a serious security risk, it just opens (for reading) and closes it.  If we want to rewrite it using stat(), that would be a separate matter (though it could be done here).  This was just supposed to address missing prototypes.

mrneont:
Hi-  In general, when a prototype is missing for a function, it would seem to make sense to resolve the situation by adding the prototype rather than by purging the function. Maintaining backwards compatibility could be considered part of reproducibility, as well?

seanm:
> It is odd that the function tries to actually open the file, rather than simply call stat()

That wouldn't solve the TOCTOU. `nifti_fileexists()` can truthfully answer yes, then an instant later some other process could delete the file. Whatever answer `nifti_fileexists()` gives you cannot be assumed to be still true one line of code later.

Such races should be discouraged, and adding this function to the public .h seems to be the opposite.

seanm:
I've reverted the -Wmissing-prototypes warning fix for `nifti_fileexists`. This way it's not "removed" from the library, but also not encouraged by begin added to the header file.

afni-rickr:
This is not a semaphore-protected read/write library, where checking is required be atomically combined with reading or writing.  The read functions have their own checks for success, as do writes.  If a file gets deleted and later operations fail, that is expected behavior.  But no big deal, you do not need to add the prototype, that can be done later.

PR #157, opened 2022-09-17
64 to 32 truncation warnings

seanm:
@afni-rickr @mrneont @hjmjohnson This PR fixes all -Wshorten-64-to-32 warnings, which is when a 64 bit integer is implicitly stored into a 32 bit integer.  Obviously this can be lossy.  I divided it into two commits, one that doesn't change any public API, and one that does.  (By public API I mean things declared in header files.)

I imagine you won't be too keen on changing the public API :), but hopefully you'll agree with the first commit at least.

### DIFF
--- a/nifti2/clib_02_nifti2.c
+++ b/nifti2/clib_02_nifti2.c
@@ -117,7 +117,7 @@ int disp_float(nifti_image * nim, int vol, int slice, int line, int offset)
 {
    float     * dp, * d2;
    long long   lloff;
-   int         nx, nxy, nxyz;
+   int64_t     nx, nxy, nxyz;
 
    if( ! nim ) return 1;
 
--- a/nifti2/nifti2_io.c
+++ b/nifti2/nifti2_io.c
@@ -496,12 +496,12 @@ static int  nifti_NBL_matches_nim(const nifti_image *nim,
                                   const nifti_brick_list *NBL);
 
 /* for nifti_read_collapsed_image: */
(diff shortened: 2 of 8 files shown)


hjmjohnson (approved):
I am supportive of all these changes.

@seanm Thank you for tracking these truncation inconsistencies down.

afni-rickr:
I think the change in the return types will crash these functions on most systems unless all such programs are remade with the updated code.  Basically, this makes the library incompatible with the previous versions (since int is almost always 32-bits).  This seems like a big deal.

seanm:
@afni-rickr I agree the API change is impactful.  But at the same time I don't think *never* changing the API is viable either, do you agree?  Assuming so, the question becomes how to manage API changes.  Does the project have a policy?

seanm:
@afni-rickr @mrneont @hjmjohnson In any case, like I said, I split this into 2 commits quite deliberately since I know the API changes will require discussion.  If you could review the actual code changes, I can make any corrections, then break this into 2 PRs.  Or would you prefer I break into 2 PRs first?

seanm:
@afni-rickr @mrneont @hjmjohnson I removed the commit (bb9dab64caf94e3e49fadc3fbbb1eb7823ede8cd) with API changes to a new PR here: https://github.com/NIFTI-Imaging/nifti_clib/pull/165

Hopefully this PR is now acceptable for merging.  It fixes, I believe, all -Wshorten-64-to-32 warnings that can be fixed without changing public API.

hjmjohnson (approved):
Thank you for your persistence @seanm .

PR #159, opened 2022-09-17
Fix reading subregions of large files

### LINKED ISSUE #158: `nifti_read_subregion_image` is limited to 2 GiB
The function [`nifti_read_subregion_image`](https://github.com/NIFTI-Imaging/nifti_clib/blob/d05a71ecdc0b811a509b77c3e82fd0674e901807/niftilib/nifti1_io.c#L6964) is limited to subregions of 2 GiB.
Bigger sizes cause a segmentation fault.

The was discovered in InsightSoftwareConsortium/ITK#3625.

---

A size bigger than the signed size of a int (2 147 483 648 bytes) causes an overflow.
https://github.com/NIFTI-Imaging/nifti_clib/blob/d05a71ecdc0b811a509b77c3e82fd0674e901807/niftilib/nifti1_io.c#L6972
The following `malloc` does not allocate any memory and thus causes a segmentation fault.
https://github.com/NIFTI-Imaging/nifti_clib/blob/d05a71ecdc0b811a509b77c3e82fd0674e901807/niftilib/nifti1_io.c#L7060

okmiim:
The type (int) of `total_alloc_size` limits the memory allocation to 2 GiB.

Fixes #158

---
~~Open problems/questions:~~
- ~~the return value is still a `int` and returns a negative value for buffers bigger than 2 GiB.
  This requires a change in signature:
  (Would break 32bit builds)~~
  1. ~~change `bytes` and return type to `long long int`~~
  2. ~~change `bytes` and return type to `int64_t`~~
 - ~~maybe also adapt `initial_offset` and `offset`?~~

### DIFF
--- a/niftilib/nifti1_io.c
+++ b/niftilib/nifti1_io.c
@@ -6969,7 +6969,7 @@ int nifti_read_subregion_image( nifti_image * nim,
   znzFile fp;                   /* file to read */
   int i,j,k,l,m,n;              /* indices for dims */
   long int bytes = 0;           /* total # bytes read */
-  int total_alloc_size;         /* size of buffer allocation */
+  size_t total_alloc_size;      /* size of buffer allocation */
   char *readptr;                /* where in *data to read next */
   int strides[7];               /* strides between dimensions */
   int collapsed_dims[8];        /* for read_collapsed_image */
@@ -7065,8 +7065,8 @@ int nifti_read_subregion_image( nifti_image * nim,
     if(g_opts.debug > 1)
       {
       fprintf(stderr,"allocation of %d bytes failed\n",total_alloc_size);
-      return -1;
       }
+    return -1;
     }
 
   /* point to start of data buffer as char * */


dzenanz (approved):

afni-rickr:
@okmiim Indeed, the nifti2 version of this function uses int64_t for those terms, and for the return value.  But the nifti1 version of the library has no instances of int64_t (or long long) at all, using size_t instead (which I guess you noticed, since you chose size_t tor total_alloc_size).
There are other instances in the nifti1 library of int being used to store products of dimensions unfortunately, which was some part of the motivation for using nifti2.

okmiim:
@afni-rickr I understand this and your suggestion it to migrate to nifti2 (presumable 64 bit only).
Besides not changing the signature, the effect of your statement regarding this change is unclear.
I see following options for this PR:
1. Do nothing
2. Prevent a segmentation fault by handling the failed malloc
3. Option 2 + change type of `total_alloc_size` to `size_t`

afni-rickr:
@okmiim I wasn't really sure of what to suggest here, since most changes will be short of a complete fix, so I was mostly just babbling.  This change looks good.

seanm:
This change should have updated the format string:

`      fprintf(stderr,"allocation of %d bytes failed\n",total_alloc_size);`

to:

`      fprintf(stderr,"allocation of %zu bytes failed\n",total_alloc_size);`

Now it will log incorrectly for large numbers.

I can fix it with other 64 bit fixes here: https://github.com/NIFTI-Imaging/nifti_clib/pull/157

Issue #162, opened 2022-11-17
Memory leak in nifti_read_subregion_image


issakomi:
S. https://github.com/InsightSoftwareConsortium/ITK/issues/3747

PR #163, opened 2022-11-17
Fixed issues in nifti_read_subregion_image

### LINKED ISSUE #161: Can `g_opts.debug > 1` change behavior
Please kindly look at the [block](https://github.com/NIFTI-Imaging/nifti_clib/blob/8f7ecce71b738746d48283400ec40f87b6d21539/niftilib/nifti1_io.c#L7120) in _nifti_read_subregion_image_

```c
              if(nread != read_amount)
                {
                if(g_opts.debug > 1)
                  {
                  fprintf(stderr,"read of %d bytes failed\n",read_amount);
                  return -1;
                  }
                }
```

It is correct that at this point the function `return -1` only if `g_opts.debug > 1`? IMHO, this may be a typo. Not sure.
It is important to clarify before testing another issue.

issakomi:
S. https://github.com/NIFTI-Imaging/nifti_clib/issues/161
Closes #161

### DIFF
--- a/nifti2/nifti2_io.c
+++ b/nifti2/nifti2_io.c
@@ -9204,11 +9204,10 @@ int64_t nifti_read_subregion_image( nifti_image * nim,
               read_amount = rs[0] * nim->nbyper; /* read a row of subregion */
               nread = nifti_read_buffer(fp, readptr, read_amount, nim);
               if(nread != read_amount) {
-                if(g_opts.debug > 1) {
+                if(g_opts.debug > 0)
                   fprintf(stderr,"read of %" PRId64 " bytes failed\n",
                           read_amount);
-                  return -1;
-                }
+                return -1;
               }
             bytes += nread;
             readptr += read_amount;
--- a/nifti2/nifti2_io.c
+++ b/nifti2/nifti2_io.c
@@ -9146,6 +9146,12 @@ int64_t nifti_read_subregion_image( nifti_image * nim,
 
   /* get the file open */
   fp = nifti_image_load_prep( nim );
+  if(znz_isnull(fp)) {
+    if(g_opts.debug > 0)
+      fprintf(stderr,"** nifti_read_subregion_image, failed load_prep\n");
+    return -1;
+  }
+
   /* the current offset is just past the nifti header, save
    * location so that SEEK_SET can be used below
    */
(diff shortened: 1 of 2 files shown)


afni-rickr:
That looks good.  Would you like to fix the similar problem under "if(*data == 0)", above in that same function?
I see those were fixed in nifti2, but not in nifti1.
Thanks!

issakomi:
Thank you!

> Would you like to fix the similar problem under "if(*data == 0)"

It is already fixed by the commit https://github.com/NIFTI-Imaging/nifti_clib/commit/8f7ecce71b738746d48283400ec40f87b6d21539

I have set the PR to draft. I am not sure what is better here to `return -1` or just warn. Maybe i`ll try to make a test and look what happens.

issakomi:
I have tested a little. Messed a [file](https://drive.google.com/file/d/1Qcxres9L_h24nX77wXS-Qudzorsu42SR/view?usp=sharing) to trigger this logic path`if(nread != read_amount)` in _nifti_read_subregion_image_. In fact if the function doesn't return at this point -- nothing happens, also _Valgrind_ doesn't detect any defects, the file is just silently read incompletely (unless _g_opts.debug >= 1_, with  _g_opts.debug >= 1_ there is a error message and function `return -1`). So there are 3 options:
- always `return -1;`
- always continue
- do nothing and close the PR

In fact this PR is just prerequisite for _znzclose_ issue #162 . The issue #162 is straightforward, because in _nifti2_io.c_ there is `znzclose(fp);` [at the end of the same _nifti_read_subregion_image_](https://github.com/NIFTI-Imaging/nifti_clib/blob/8f7ecce71b738746d48283400ec40f87b6d21539/nifti2/nifti2_io.c#L9222) function.

afni-rickr:
Hi @issakomi.  This function is used by a different group, I believe (though I did put a call in nifti_tool, for code coverage and regression testing), and they likely trap on < 0.  Since the intention was to fail on a short read, and since this only affects an error condition, return -1 does seem like the way to go.
On top of that, trapping a bad return from nifti_image_load_prep() and closing fp seem great to do, if you would like to include them.  Thanks a lot!

issakomi:
Thank you!

> On top of that, trapping a bad return from nifti_image_load_prep() and closing fp seem great to do, if you would like to include them.

I have done this in separate commit.  This will close the issues  #162 too. If you like can merge both commit into one.

afni-rickr:
This looks very good to me as is, thanks!  @hjmjohnson, do you have any comments?

hjmjohnson (approved):

PR #165, opened 2022-12-29
64 to 32 truncation warnings api changes

seanm:


seanm:
As this is a public API change, I'm not pushing for it it be merged ASAP.

But at some point, at a major version bump, it would be nice to have these integrated.

PR #166, opened 2023-01-03
Fixed various type punning warnings by using memcpy()

seanm:
Modern compilers optimize memcpy() very well, this will likely produce the same code.

### DIFF
--- a/nifti2/nifti_tool.c
+++ b/nifti2/nifti_tool.c
@@ -3822,7 +3822,6 @@ int modify_all_fields( void * basep, nt_opts * opts, field_s * fields, int flen)
  *----------------------------------------------------------------------*/
 int modify_field(void * basep, field_s * field, const char * data)
 {
-   float   fval;
    const char  * posn = data;
    int     val, max, fc, nchars;
 
@@ -3849,7 +3848,7 @@ int modify_field(void * basep, field_s * field, const char * data)
 
          case DT_INT8:
          {
-            max = 127;
+            max = INT8_MAX;
             for( fc = 0; fc < field->len; fc++ )
             {
                if( sscanf(posn, " %d%n", &val, &nchars) != 1 )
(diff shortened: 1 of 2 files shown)


nifti2/nifti_tool.c
```diff
@@ -3849,7 +3848,7 @@ int modify_field(void * basep, field_s * field, const char * data)
 
          case DT_INT8:
          {
-            max = 127;
+            max = UINT8_MAX;
```
issakomi, line 3851: Is this change intentional? 127 vs. 255. Also 32767 vs. 65535 below.

seanm: Whoops! Excellent catch.

seanm: Fixed.

hjmjohnson (approved):

seanm:
Thanks @hjmjohnson !

afni-rickr:
Note that this commit has been undone with PR 170.  The addressing in 166 does not account for types in indexing.  For example `char * offsetp = &(((char *)basep + field->offset)[fc]);` with float data needs [fc] to offset by floats, not chars.

seanm:
> Note that this commit has been undone with PR 170. The addressing in 166 does not account for types in indexing.

Oh damn, indeed you are right.  The many brackets threw me for a loop.  Disturbing that no test failed due to this.  I'll make a new PR to fix it.

afni-rickr:
I could add a test for that, something like:
```
nifti_tool -mod_hdr -new_dim 4 64 64 21 180 0 0 0                \
            -mod_field pixdim "0.0 4.0 4.0 6.0 3.0 1.0 1.0 1.0"  \
            -prefix new_image.nii -infiles MAKE_IM
nifti_tool -disp_hdr -field pixdim -quiet -infiles new_image.nii
```
If pixdim were set using a variable, that could be checked against the output.

seanm:
@afni-rickr please see my fix in https://github.com/NIFTI-Imaging/nifti_clib/pull/172

PR #169, opened 2023-03-31
Increased an integer to long to fix signed integer overflow

seanm:
With the ITK unit test itkNiftiLargeImageRegionReadTest, UBSan correctly reported:

runtime error: signed integer overflow: 1020 * 2138312 cannot be represented in type 'int'

Luckily, the private variable `strides` can just be made long.

### DIFF
--- a/niftilib/nifti1_io.c
+++ b/niftilib/nifti1_io.c
@@ -6928,7 +6928,7 @@ int nifti_read_collapsed_image( nifti_image * nim, const int dims [8],
 ** stride array.
 */
 static void
-compute_strides(int *strides,const int *size,int nbyper)
+compute_strides(int64_t *strides,const int *size,int nbyper)
 {
   int i;
   strides[0] = nbyper;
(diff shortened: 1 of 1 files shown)


seanm:
@hjmjohnson care to review?

afni-rickr:
please do not merge anything right now

seanm:
I've just remembers/realized this code also exists in the nifti2 folder, so I compared them, and there `int64_t` is used, so I've changed my use of `long` to `int64_t` to match.

seanm:
@afni-rickr this should be quick to review and safe to merge, if you have some time to look...

afni-rickr:
This look okay to me, though it introduces int64_t to the niftilib directory.  The NIFTI-1 implementation came out in 2003, so C99 was quite new then, and was certainly not required.
@hjmjohnson, this is your code, what do you prefer?

seanm:
@hjmjohnson let us know if you prefer `long long` (C89 compatible) over `int64_t` (matching nifti2 folder).  If the latter, I'll amend this to also `#include <stdint.h>`

seanm:
@hjmjohnson friendly ping...

hjmjohnson (approved):

PR #171, opened 2023-04-01
some updates for PR 157

afni-rickr:
- put variable definitions at function blocks
- print int64_t using PRId64
- znz_off_t varies, so try not to print
- extensions are int size, limit size and check in read_file_text()
- call strlen() only after NULL check
- modify_field(), check length before copy
- act_disp_ci() limited to int number of elements
- formatting

### DIFF
--- a/nifti2/nifti2_io.c
+++ b/nifti2/nifti2_io.c
@@ -4960,6 +4960,7 @@ nifti_image* nifti_convert_n2hdr2nim(nifti_2_header nhdr, const char * fname)
 {
    int64_t      ii;
    int          doswap, ni_ver, is_onefile;
+   int          byteOrder;
    nifti_image *nim;
 
    nim = (nifti_image *)calloc( 1 , sizeof(nifti_image) ) ;
@@ -5021,7 +5022,7 @@ nifti_image* nifti_convert_n2hdr2nim(nifti_2_header nhdr, const char * fname)
 
    nim->nifti_type = (is_onefile) ? NIFTI_FTYPE_NIFTI2_1 : NIFTI_FTYPE_NIFTI2_2;
 
-   int byteOrder = nifti_short_order() ;
+   byteOrder = nifti_short_order() ;
    if( doswap )   nim->byteorder = REVERSE_ORDER(byteOrder) ;
    else           nim->byteorder = byteOrder ;
 
@@ -6084,6 +6085,7 @@ nifti_image * nifti_read_ascii_image(znzFile fp, const char *fname, int flen,
                                      int read_data)
 {
    nifti_image * nim;
+   int64_t       slen;
    int           txt_size, remain, rv = 0;
    char        * sbuf, lfunc[25] = { "nifti_read_ascii_image" };
 
@@ -6092,11 +6094,12 @@ nifti_image * nifti_read_ascii_image(znzFile fp, const char *fname, int flen,
               fname);
      return NULL;
    }
-   int64_t slen = flen;  /* slen will be our buffer length */
+   slen = flen;  /* slen will be our buffer length */
    if( slen <= 0 ) slen = nifti_get_filesize(fname);
 
    if( g_opts.debug > 1 )
-      fprintf(stderr,"-d %s: have ASCII NIFTI file of size %lld\n",fname,slen);
+      fprintf(stderr,"-d %s: have ASCII NIFTI file of size %" PRId64 "\n",
+              fname, slen);
 
    if( slen > 65530 ) slen = 65530 ;
    sbuf = (char *)calloc(sizeof(char),slen+1) ;
@@ -9258,7 +9261,8 @@ static int rci_read_data(nifti_image * nim, int64_t * pivots, int64_t * prods,
 
       /* make sure things look good here */
       if( *pivots != 0 ){
-         fprintf(stderr,"** NIFTI rciRD: final pivot == %lld!\n", *pivots);
+         fprintf(stderr,"** NIFTI rciRD: final pivot == %" PRId64 "\n",
+                 *pivots);
          return -1;
       }
 
@@ -9361,8 +9365,8 @@ static int rci_alloc_mem(void **data, const int64_t prods[8], int nprods, int nb
    wants to collapse a dimension.  The last pivot should always be zero
    (note that we have space for that in the lists).
 */
-static int make_pivot_list(nifti_image *nim, const int64_t dims[], int64_t pivots[],
-                                             int64_t prods[], int * nprods )
+static int make_pivot_list(nifti_image *nim, const int64_t dims[],
+                           int64_t pivots[], int64_t prods[], int * nprods )
 {
    int len = 0;
    int64_t dind = nim->dim[0];
@@ -9389,7 +9393,7 @@ static int make_pivot_list(nifti_image *nim, const int64_t dims[], int64_t pivot
    if( g_opts.debug > 2 ){
       fprintf(stderr,"+d pivot list created, pivots :");
       for(dind = 0; dind < len; dind++)
-         fprintf(stderr," %lld", pivots[dind]);
+         fprintf(stderr," %" PRId64 "", pivots[dind]);
       fprintf(stderr,", prods :");
       for(dind = 0; dind < len; dind++)
          fprintf(stderr," %" PRId64 "", prods[dind]);
--- a/nifti2/nifti_tool.c
+++ b/nifti2/nifti_tool.c
@@ -208,7 +208,6 @@ static char g_version[] = "2.13";
 static char g_version_date[] = "February 27, 2022";
 static int  g_debug = 1;
 
-#include <errno.h>
 #include <limits.h>
 #include <string.h>
 #include <stdint.h>
@@ -219,7 +218,7 @@ static int  g_debug = 1;
 /* local prototypes */
 static int free_opts_mem(nt_opts * nopt);
 static int num_volumes(nifti_image * nim);
-static char * read_file_text(const char * filename, int64_t * length);
+static char * read_file_text(const char * filename, int * length);
 
 
 #define NTL_FERR(func,msg,file)                                      \
@@ -402,12 +401,12 @@ int process_opts( int argc, char * argv[], nt_opts * opts )
       /* begin normal execution options... */
       else if( ! strcmp(argv[ac], "-add_ext") )
       {
+         int64_t new_ecode;
          ac++;
          CHECK_NEXT_OPT(ac, argc, "-add_ext");
-         errno = 0;
-         long long new_ecode = strtoll(argv[ac], NULL, 10);
-         if ((new_ecode == 0 && errno != 0) || new_ecode <= INT_MIN || new_ecode >= INT_MAX) {
-           fprintf(stderr,"** invalid -add_ext parameter\n");
+         new_ecode = strtol(argv[ac], NULL, 10);
+         if (new_ecode <= 0 || new_ecode > INT_MAX) {
+           fprintf(stderr,"** invalid -add_ext code, %s\n", argv[ac]);
            return -1;
          }
          if( add_int(&opts->etypes, (int)new_ecode) ) return -1;
@@ -847,17 +846,19 @@ int verify_opts( nt_opts * opts, char * prog )
 
 /*----------------------------------------------------------------------
  * re-assemble the command string into opts->command
+ * (commands are currently limited to 2048 bytes)
  *----------------------------------------------------------------------*/
 int fill_cmd_string( nt_opts * opts, int argc, char * argv[])
 {
    char * cp;
-   size_t remain = sizeof(opts->command);  /* max command len */
+   int    len, remain = (int)sizeof(opts->command);  /* max command len */
    int    c, ac;
    int    has_space;  /* arguments containing space must be quoted */
    int    skip = 0;   /* counter to skip some of the arguments     */
 
    /* get the first argument separately */
-   int len = snprintf( opts->command, sizeof(opts->command), "\n  command: %s", argv[0] );
+   len = snprintf( opts->command, sizeof(opts->command),
+                   "\n  command: %s", argv[0] );
    if( len < 0 || len >= (int)sizeof(opts->command) ) {
       fprintf(stderr,"FCS: no space remaining for command, continuing...\n");
       return 1;
@@ -870,7 +871,7 @@ int fill_cmd_string( nt_opts * opts, int argc, char * argv[])
    {
       if( skip ){ skip--;  continue; }  /* then skip these arguments */
 
-      size_t len = strlen(argv[ac]);
+      len = (int)strlen(argv[ac]);
       if( len + 3 >= remain ) {  /* extra 3 for space and possible '' */
          fprintf(stderr,"FCS: no space remaining for command, continuing...\n");
          return 1;
@@ -2239,8 +2240,7 @@ int act_add_exts( nt_opts * opts )
    nifti_image      * nim;
    const char       * ext;
    char             * edata = NULL;
-   int                fc, ec;
-   int64_t            elen;
+   int                fc, ec, elen;
 
    if( g_debug > 2 ){
       fprintf(stderr,"+d adding %d extensions to %d files...\n"
@@ -2264,10 +2264,10 @@ int act_add_exts( nt_opts * opts )
 
       for( ec = 0; ec < opts->elist.len; ec++ ){
          ext = opts->elist.list[ec];
-         elen = strlen(ext);
+         elen = (int)strlen(ext);
          if( !strncmp(ext,"file:",5) ){
             edata = read_file_text(ext+5, &elen);
-            if( !edata || elen <= 0 || elen > INT_MAX) {
+            if( !edata || elen <= 0 ) {
                fprintf(stderr,"** failed to read extension data from '%s'\n",
                        ext+5);
                continue;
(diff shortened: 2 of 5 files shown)


nifti2/nifti2_io.c
```diff
@@ -5021,7 +5022,7 @@ nifti_image* nifti_convert_n2hdr2nim(nifti_2_header nhdr, const char * fname)
 
    nim->nifti_type = (is_onefile) ? NIFTI_FTYPE_NIFTI2_1 : NIFTI_FTYPE_NIFTI2_2;
 
-   int byteOrder = nifti_short_order() ;
+   byteOrder = nifti_short_order() ;
```
seanm, line 5025: Why split the declaration and initial assignment?  The codebase clearly doesn't support C89, as things like `int64_t` are used unconditionally.

nifti2/nifti_tool.c
```diff
@@ -402,15 +401,15 @@ int process_opts( int argc, char * argv[], nt_opts * opts )
...
       else if( ! strcmp(argv[ac], "-add_ext") )
       {
+         int new_ecode;
          ac++;
          CHECK_NEXT_OPT(ac, argc, "-add_ext");
-         errno = 0;
-         long long new_ecode = strtoll(argv[ac], NULL, 10);
-         if ((new_ecode == 0 && errno != 0) || new_ecode <= INT_MIN || new_ecode >= INT_MAX) {
-           fprintf(stderr,"** invalid -add_ext parameter\n");
+         new_ecode = atoi(argv[ac]);
```
seanm, line 407: Why revert to atoi?  It is obsolete. It results in undefined behaviour if the string can't fit into an int.

See for example:

https://wiki.sei.cmu.edu/confluence/display/c/MSC24-C.+Do+not+use+deprecated+or+obsolescent+functions

nifti2/nifti_tool.c
```diff
@@ -2323,20 +2323,25 @@ int act_add_exts( nt_opts * opts )
...
       return NULL;
    }
 
-   int64_t len = nifti_get_filesize(filename);
-   if( len <= 0 ) {
+   len64 = nifti_get_filesize(filename);
+   if( len64 <= 0 ) {
       fprintf(stderr,"** RFT: file '%s' appears empty\n", filename);
       return NULL;
+   } else if( len64 > INT_MAX ) {
```
seanm, line 2342: INT32_MAX would be more correct.

nifti2/nifti_tool.c
```diff
@@ -3829,8 +3834,7 @@ int modify_field(void * basep, field_s * field, const char * data)
    if( g_debug > 1 )
       fprintf(stderr,"+d modifying field '%s' with '%s'\n", field->name, data);
 
-   size_t dataLength = strlen(data);
-   if( !data || dataLength == 0 )
+   if( !data || strlen(data) == 0 )
```
seanm, line 3837: good catch!

nifti2/nifti_tool.c
```diff
@@ -6689,12 +6695,17 @@ int act_disp_ci( nt_opts * opts )
...
 
-      int64_t len = nifti_read_collapsed_image(nim, opts->ci_dims, &data);
-      if( len < 0 || !data )
+      len64 = nifti_read_collapsed_image(nim, opts->ci_dims, &data);
+      if( len64 < 0 || !data )
       {
          fprintf(stderr,"** FAILURE for dataset '%s'\n", nim->fname);
          if( data ) { free(data); data = NULL; }
          err++;
+      } else if ( len64/nim->nbyper > INT_MAX ) {
```
seanm, line 6704: INT32_MAX would be more correct.

nifti2/nifti_tool.c
```diff
@@ -6689,12 +6695,17 @@ int act_disp_ci( nt_opts * opts )
...
+         fprintf(stderr,"** %" PRId64 " is too many values to display\n",
+                 len64/nim->nbyper);
+         if( data ) { free(data); data = NULL; }
```
seanm, line 6707: The `if(data)` check is unnecessary. `free(NULL)` is well-defined.

nifti2/nifti_tool.c
```diff
@@ -6710,7 +6721,8 @@ int act_disp_ci( nt_opts * opts )
          fprintf(stdout,")\n");
       }
 
-      disp_raw_data(data, nim->datatype, len / nim->nbyper, space, 1);
+      /* should we change disp_raw_data to allow for 64-bit nvalues? */
```
seanm, line 6724: We should, and I did in https://github.com/NIFTI-Imaging/nifti_clib/pull/165

nifti2/nifti2_io.c
afni-rickr: While not required by the compiler, it is certainly the convention in the code, making location of types predictable.

nifti2/nifti2_io.c
```diff
@@ -9258,7 +9261,8 @@ static int rci_read_data(nifti_image * nim, int64_t * pivots, int64_t * prods,
 
       /* make sure things look good here */
       if( *pivots != 0 ){
-         fprintf(stderr,"** NIFTI rciRD: final pivot == %lld!\n", *pivots);
+         fprintf(stderr,"** NIFTI rciRD: final pivot == %" PRId64 "\n",
```
seanm, line 9264: Did you mean to remove the exclamation point?

nifti2/nifti_tool.c
afni-rickr: Hmmm, it looks like it is being pushed towards obsolescence, but of course, will never be.  But I believe it returns 0 on error, not an undefined value.  I wanted to remove the unnecessary use of long long (which is a very annoying type, since it's actual type varies).  The ecode is question only has a small range of acceptable values, which will be checked.  Using atoi() is not important, but going to large types is ugly.

nifti2/nifti_tool.c
afni-rickr: Sure.  I almost never use those terms, but would expect INT_MAX to be reasonably appropriate when comparing with ints.  Either is fine.  We require int to mean int32_t.

nifti2/nifti2_io.c
seanm: OK, if that's the project's style, I can follow it in the future, and update my other pending PRs accordingly.

But I don't think that style is advantageous.  It's not for nothing that C99 and C++ got rid of the old C requirement for having declarations at the top.  It puts everything in a larger scope than necessary, see for example:

https://wiki.sei.cmu.edu/confluence/display/c/DCL19-C.+Minimize+the+scope+of+variables+and+functions

Tools like cppcheck and clang-tidy can warn about such things. I was going to work on such a scope reduction patch after my other cppcheck PR lands.

nifti2/nifti_tool.c
seanm: >But I believe it returns 0 on error, not an undefined value

[C99 7.20.1](http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1256.pdf) "If the value of the result cannot be represented, the behavior is undefined."

> ... use of long long (which is a very annoying type, since it's actual type varies)

That's true of all the types though, even the `int` that `atoi` uses.

>but going to large types is ugly

Ugly is preferable to undefined behaviour though.

nifti2/nifti_tool.c
afni-rickr: Oh right, though the counter was only of type int.
Still, it would be more clear to put functionality enhancement PRs separate from code cleanup ones.  This looked more like "well, we've altered the type, so now it should propagate".  It would be fine to go 64-bit here, but if you want that, maybe just make it a separate change.

nifti2/nifti2_io.c
afni-rickr: Yes, it seemed like a mistake to be there.  I am not a fan of programs yelling at the users.  :)

afni-rickr: BTW, way to pay attention to the little details...  :)

niftilib/nifti1_io.c
```diff
@@ -4405,15 +4405,15 @@ static int nifti_read_extensions( nifti_image *nim, znzFile fp, int remain )
       return -1;
    }
 
-   znz_off_t posn = znztell(fp);
+   posn = (int)znztell(fp);
```
seanm, line 4408: `znztell` returns `znz_off_t`.  The cast you've added could cause information loss.

You mentioned "znz_off_t varies, so try not to print" which I guess is why you made this change...

But what's wrong with using `%lld` for `znz_off_t`?  If `znz_off_t` is smaller than `long long` it'll still get printed correctly.

With this change, if `znztell` returns a huge number, your error message won't even print the actual thing `znztell` returned.

nifti2/nifti_tool.c
seanm: >Still, it would be more clear to put functionality enhancement PRs separate from code cleanup ones.

I did split my `-Wshorten-64-to-32` fixes into 2 PRs:

- #157 which changed no APIs, that this PR #171 is partly reverting
- #165 which does change APIs like `disp_raw_data` above.  It's awaiting review.

niftilib/nifti1_io.c
afni-rickr: posn is required to be 358, anything else is a fatal error.  And %lld will produce compiler warnings on many systems (since znz_off_t varies).  Using int64_t would be okay, but seems like overkill.

afni-rickr: Oh wait, it is not a fatal error, just a warning (since we don't care soooo much about extensions, I guess)...

seanm: > And %lld will produce compiler warnings on many systems (since znz_off_t varies)

Ah true, -Wformat will warn about that. Then the most correct thing is to:

1. not do the truncating cast you added
2. cast the %lld param to `long long` to match `lld`

```c
znz_off_t posn = znztell(fp);

fprintf(stderr,"-d nre: posn = %lld, offset = %d, type = %d, remain = %d\n",
    (long long)posn, nim->iname_offset, nim->nifti_type, remain);
```

niftilib/nifti1_tool.c
```diff
@@ -3915,10 +3919,11 @@ int disp_raw_data( void * data, int type, int nvals, char space, int newline )
 int clear_float_zeros( char * str )
 {
    char * dp  = strchr(str, '.'), * valp;
+   int    len;
 
    if( !dp ) return 0;      /* nothing to clear */
 
-   size_t len = strlen(dp);
+   len = (int)strlen(dp);
```
seanm, line 3926: Why not use `size_t` here?

Sure, it's super-unlikely that any string passed to this function will be longer than INT_MAX, but still... are you trying to save 4 bytes of stack space?

afni-rickr: The only reason size_t was added was because of warnings, and casting is generally the more simple and less impactful way to do that.  The function is clearing some number of zeros from a maximum 32 character string.  Unnecessarily promoting variables to higher sizes confuses the purposes of the functions.  int isn't better than size_t here, but it would have been the preferred change in the first place.  This is just more in line with how overall changes should be made to this mature (but not perfect) library.  Unless something is actually broken, have as little impact as possible.

seanm: >The only reason size_t was added was because of warnings

I turned on `-Wshorten-64-to-32` *to find bugs*, not just to shut the compiler up.  This function `clear_float_zeros` is buggy. It does not do what it's supposed to do when the string is huge. The warning found this bug for us.

>Unnecessarily promoting variables to higher sizes...

There is no promotion to higher sizes.  `strlen` itself returns `size_t`.  `strlen` will correctly compute the length of a huge string.  My changing the `len` variable from `int` to `size_t` only makes it match what `strlen` is *already* giving us. Your changing it back to `int` here is discarding 50% of `strlen`'s result.

We can even [compare the assembly](https://godbolt.org/z/Y4Mradjhe) of the `int` vs `size_t` variants. The truncation to int *takes extra work*, on top of just being incorrect.

>Unless something is actually broken, have as little impact as possible.

I understand not wanting to add regressions to a mature codebase. We indeed have to be careful of that. But something *is* actually broken here, albeit not something likely to be encountered in normal use.

The fix (changing `int` to `size_t`) is extremely safe.

niftilib/nifti1_tool.c
```diff
@@ -3885,7 +3885,9 @@ int disp_raw_data( void * data, int type, int nvals, char space, int newline )
          {
                float temp;
                memcpy(&temp, dp, sizeof(temp));
-               snprintf(fbuf,sizeof(fbuf),"%f", temp);
+               snprintf(fbuf, NT_LOC_MAX_FLOAT_BUF,"%f", temp);
+               /* terminate, since some versions might not */
```
seanm, line 3889: I've never heard of any `snprintf` with such a disastrous bug...

Are you perhaps thinking of the old Microsoft `_snprintf` (note the leading underscore)?  See:

https://stackoverflow.com/questions/7706936/is-snprintf-always-null-terminating

As we are not calling `_snprintf` this change should not be needed.

niftilib/nifti1_tool.c
seanm: Thank you for bring this back to `size_t`. I appreciate you being open to argument!

nifti2/nifti_tool.c
afni-rickr: If you would like to make the 64-bit update for this, please do not change library APIs in that PR.  Modifying types in nifti_tool is okay, modifying them in the library is a bigger deal.  Let's keep them separate.  Any library updates should be done in the context of heading to a 4.x.x release.  We should make useful updates in the current 3.x.x version, first.

niftilib/nifti1_tool.c
afni-rickr: "Bug" is not for us to decide regarding this, the question is what does the doc say vs what does the function do.  Reading the man page many times, it is not 100% clear that it always does the termination, the help seems a bit ambiguous.  But yes, I guess it was the _ version that raised the flag.
So I don't really care about this either way.  Being safe is good, not mentioning M!cr*s@ft is very good.

nifti2/nifti2_io.c
afni-rickr: Yes, please do not alter the style of programming in this library.  Save that for a new library if you wish to.

niftilib/nifti1_io.c
afni-rickr: Okay, will do.  I sill worry about getting warnings, we will see.

nifti2/nifti2_io.c
seanm: OK, will keep the C89 style in future changes.

niftilib/nifti1_io.c
seanm: `-Wformat` won't warn then, I tried.

niftilib/nifti1_tool.c
seanm: >"Bug" is not for us to decide regarding this

The C standard requires snprintf to ***always*** null terminate. I've updated sprintf to snprintf in many open source libraries recently, and have never seen the MS version having any problem.  It was only the old non-standard underscore version that was busted.

I would remove this change, it an unnecessary complication.

afni-rickr: Okay, this is the last question remaining for this PR.  You can choose: keep the termination or remove it, what do you prefer?  I am happy keeping it.  If I have to do google searches to verify the functionality, it is preferable not to make "the next programmer" have to as well, so clarity is nice.

afni-rickr: Okay, I will remove this, thanks.

seanm:
@afni-rickr can you squash the 3 commits?  Otherwise history will be weird with changes done and undone.

niftilib/nifti1_io.c
afni-rickr: These things can vary across systems, as the sizes vary.  "long" types do not seem to be so reliable.

afni-rickr:
I was going to just do this as a new PR and a single commit, unless you know of a sneakier way to go.

seanm:
> I was going to just do this as a new PR and a single commit, unless you know of a sneakier way to go.

You can transform this PR into a single commit by "squashing the commits".  Is that something you are unfamiliar with?  If so, it's not that hard, I can walk you through it...

niftilib/nifti1_io.c
seanm: Their size varies, but "%lld" means "long long" regardless of what the size of long long is.

niftilib/nifti1_tool.c
afni-rickr: BTW, I did look at the C standard, and indeed it is quite clear about the null termination, while the man page is not quite so clear.  That should by my go-to reference, rather than the man pages.  Thanks.

seanm:
I would do `git rebase -i HEAD~3` and mark the last 2 commits to be squashed.  Then `git push -f origin HEAD` to update this PR.  The force is no problem because this PR isn't merged to master yet, so nothing else is based on these commits.

afni-rickr:
Okay, will do.  But to be sure, this would cause your local branch of update_157 to diverge, is that right.
Thanks for the info.

seanm:
> Okay, will do. But to be sure, this would cause your local branch of update_157 to diverge, is that right? Thanks for the info.

I only have a local branch to do code review on my own computer, for big changes I find the github UI not great.

seanm:
Ah, there were a few other changes I was going to suggest, but as you've already merged, I'll make a new PR...

PR #172, opened 2023-04-01
Fixed various type punning warnings by using memcpy()

seanm:
2acc4449 was buggy, and undone with cedc42d2.  This essentially restores 2acc4449 without the buggy pointer arithmetic.

Modern compilers optimize memcpy() very well, this will likely produce the same code.

### DIFF
--- a/nifti2/nifti_tool.c
+++ b/nifti2/nifti_tool.c
@@ -211,6 +211,7 @@ static int  g_debug = 1;
 #include <limits.h>
 #include <string.h>
 #include <stdint.h>
+#include <assert.h> // TEMP
 
 #include "nifti2_io.h"
 #include "nifti_tool.h"
@@ -3828,7 +3829,6 @@ int modify_all_fields( void * basep, nt_opts * opts, field_s * fields, int flen)
  *----------------------------------------------------------------------*/
 int modify_field(void * basep, field_s * field, const char * data)
 {
-   float         fval;
    const char  * posn = data;
    int           val, max, fc, nchars;
    size_t        dataLength;
@@ -3856,7 +3856,7 @@ int modify_field(void * basep, field_s * field, const char * data)
 
          case DT_INT8:
          {
-            max = 127;
+            max = INT8_MAX;
             for( fc = 0; fc < field->len; fc++ )
             {
                if( sscanf(posn, " %d%n", &val, &nchars) != 1 )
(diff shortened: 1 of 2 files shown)


seanm:
@afni-rickr @hjmjohnson this is a fixed version of PR #166 which was buggy and undone by PR #170.

Don't merge as-is, I added some asserts to verify the address calculation is correct. They are just there as a test and will be removed.

Alas, no test actually enters this code, as I confirmed with the `assert(0)`.

nifti2/nifti_tool.c
```diff
@@ -212,6 +212,7 @@ static int  g_debug = 1;
 #include <limits.h>
 #include <string.h>
 #include <stdint.h>
+#include <assert.h> // TEMP
```
hjmjohnson, line 215: Should this //TEP be removed before committing?

nifti2/nifti_tool.c
```diff
@@ -3894,7 +3895,12 @@ int modify_field(void * basep, field_s * field, const char * data)
                   return 1;
                }
                /* otherwise, we're good */
-               ((short *)((char *)basep + field->offset))[fc] = (short)val;
+               int16_t sval = (int16_t)val;
+               char * offsetp = (char *)basep + field->offset;
+               offsetp = &(offsetp[fc * sizeof(sval)]);
+               assert(offsetp == &(((short *)((char *)basep + field->offset))[fc])); // TEMP
+               assert(0);
```
hjmjohnson, line 3902: Can this assert(0) be removed?

hjmjohnson (changes requested):

nifti2/nifti_tool.c
afni-rickr: Are the other assert()s going to stay?  If so, that means some compiler warns about the original statements but not the asserts?
If the asserts will be kept, they should probably have casts to (char *)before the '&'.

afni-rickr: Also, going from a pointer to an array and back seems a little odd.  What if 3900 were to read:
   `offsetp += fc*sizeof(sval);  /* offset further by fc elements */`

Or that could be merged into the previous line, though it is nice to spell out the pieces:
```
    /* from the struct base offset by the field location, access element #fc */
    char * offsetp = (char *)basep + field->offset + fc*sizeof(sval);
```

seanm: Please see my comments way above, from 2023-04-01, repeated here:

"Don't merge as-is, I added some asserts to verify the address calculation is correct. They are just there as a test and will be removed.

Alas, no test actually enters this code, as I confirmed with the assert(0)."

It would be good if we first added some tests that exercise this code. Then we can confirm the code change is correct by not hitting those asserts.

PR #174, opened 2023-07-17
Removed unnecessary guards against free(NULL)

seanm:
free()ing NULL has been well-defined to do nothing since C89.

This simplifies the code with no behaviour change.


seanm:
@afni-rickr @hjmjohnson friendly ping. Care to review?

hjmjohnson (approved):

PR #176, opened 2023-09-12
Eliminated duplicate strlen() call, fix strlen() truncation to int

seanm:
strlen() is O(N) and thus slow for long strings. modify_field() always calls strlen() at the beginning, but was sometimes calling it again later. No need to compute the length twice.

Also use size_t for the return value of strlen() instead of truncating to int.


seanm:
@afni-rickr @hjmjohnson care to review?

hjmjohnson (approved):

afni-rickr:
Let's please try to restrict changes to those that are useful.  I would not approve this PR.

PR #179, opened 2023-10-13
Make opts const, allowing literal parameter.

shangjiaxuan:
The following code currently fails compile because function parameter `opts` is declared as non-const, disallowing literal constants.

```
	nifti_image* img = NULL;
	znzFile file = nifti_image_open("image.nii", "rb", &img);
```

### DIFF
--- a/nifti2/nifti2_io.c
+++ b/nifti2/nifti2_io.c
@@ -5248,7 +5248,7 @@ nifti_image* nifti_convert_n2hdr2nim(nifti_2_header nhdr, const char * fname)
         <br>NULL if something fails badly.
     \sa nifti_image_load, nifti_image_free
  */
-znzFile nifti_image_open(const char * hname, char * opts, nifti_image ** nim)
+znzFile nifti_image_open(const char * hname, const char * opts, nifti_image ** nim)
 {
   znzFile fptr=NULL;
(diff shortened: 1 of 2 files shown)


seanm:
@shangjiaxuan this would be a change to public API and so needs to be done carefully.  The maintainers are conservative about changing public API.  It might be best to do an audit of where `const` is missing and batch all such changes together. Similar to https://github.com/NIFTI-Imaging/nifti_clib/pull/165

shangjiaxuan:
@seanm after searchin through the source code, I believe this is the only function that needs `const` added. In fact the corresponding function in `nifti1_io.h` and `nifti1_io.c` is declared and implemented as `const`. The use of `opts` is for the `fopen("file", "rb")`'s "rb" mode parameter, and literal is expected on same code path.

seanm:
@shangjiaxuan oh, indeed that is weird that it ***is*** const in nifti1_io.c but *not* in nifti2_io.c.  I would have expected the opposite.

@hjmjohnson @afni-rickr given the above, I vote for merging this. What do you think?

afni-rickr:
I do not see a compilation failure for that, so it presumably depends on build options.  But this seems like a fine update.
Thanks @shangjiaxuan.

Issue #184, opened 2024-01-11
Compilation fails in Visual Studio: missing stdint.h include


fdgStilla:
`nifti1_io.c` is using `int64_t` which is defined in `stdint.h` according to https://learn.microsoft.com/en-us/cpp/c-runtime-library/standard-types?view=msvc-160 but the header is not included

Found while trying to compile ITK 5.4 on windows

fdgStilla:
For info, in the default ITK build `stdint.h` is provided by chance by `Modules\ThirdParty\ZLIB\src\itkzlib-ng\zlib.h`, I saw the issue when trying to compile with conan which replace the itk zlib implementation by another.

PR #185, opened 2024-01-11
Add missing header

fdgStilla:
Fix https://github.com/NIFTI-Imaging/nifti_clib/issues/184

### DIFF
--- a/niftilib/nifti1_io.c
+++ b/niftilib/nifti1_io.c
@@ -1,5 +1,6 @@
 #define NIFTI1_IO_C
 
+#include <stdint.h>
 #include <assert.h>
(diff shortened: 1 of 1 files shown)


seanm:
This was probably my fault from https://github.com/NIFTI-Imaging/nifti_clib/pull/169

After looking a bit... nifti2_io.c doesn't have this problem because nifti2.h includes `stdint.h`. Maybe for symmetry we should instead fix this by making nifti1.h also include `stdint.h`?

@hjmjohnson @afni-rickr thoughts?

seanm:
If I could merge, I would.  I also have many PRs waiting.

@hjmjohnson @afni-rickr friendly ping...

PR #187, opened 2024-05-16
Add support for building as shared library in msvc

Julianiolo:
Hi :)

This is a "fix" to support building the libraries as shared libraries on msvc. 
This was previously not possible, because msvc requres symbols to be marked for exporting.
Cmake has a "dirty" fix for this: [CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS](https://cmake.org/cmake/help/latest/variable/CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS.html), however this might export too much and also does not work for global variables (such as `inam` in `nifticdf`).

So I added `..._API` defines which turn into either `__declspec( dllexport )` or `__declspec( dllimport )` when building shared libs. 

Let me know what you think of this :)

### DIFF
--- a/cifti/afni_xml_io.h
+++ b/cifti/afni_xml_io.h
@@ -51,25 +51,27 @@
  * ----------------------------------------------------------------------*/
 
 #ifndef CIF_API
-  #if defined(_WIN32) || defined(__CYGWIN__)
-    #if defined(CIFTI_BUILD_SHARED)
+   #if defined(_WIN32) || defined(__CYGWIN__)
+      #if defined(CIFTI_BUILD_SHARED)
       #ifdef __GNUC__
-        #define CIF_API __attribute__ ((dllexport))
+         #define CIF_API __attribute__ ((dllexport))
       #else
-        #define CIF_API __declspec((dllexport))
+         #define CIF_API __declspec( dllexport )
       #endif
-    #elif defined(CIFTI_USE_SHARED)
+      #elif defined(CIFTI_USE_SHARED)
       #ifdef __GNUC__
-        #define CIF_API __attribute__ ((dllimport))
+         #define CIF_API __attribute__ ((dllimport))
       #else
-        #define CIF_API __declspec((dllimport))
+         #define CIF_API __declspec( dllimport )
       #endif
-    #endif
-  #elif (defined(__GNUC__) && __GNUC__ >= 4) || defined(__clang__)
-    #define CIF_API __attribute__ ((visibility ("default")))
-  #else
-    #define CIF_API
-  #endif
+      #else
+      #define CIF_API
+      #endif
+   #elif (defined(__GNUC__) && __GNUC__ >= 4) || defined(__clang__)
+      #define CIF_API __attribute__ ((visibility ("default")))
+   #else
+      #define CIF_API
+   #endif
 #endif
 
 /* --------------------------- structures --------------------------------- */
--- a/cifti/afni_xml_io.h
+++ b/cifti/afni_xml_io.h
@@ -5,6 +5,16 @@
 #include <stdio.h>
 #include "afni_xml.h"
 
+#if !defined(CIF_API) && defined(_WIN32) 
+#if defined(CIFTI_BUILD_SHARED)
+#define CIF_API __declspec( dllexport )
+#elif defined(CIFTI_USE_SHARED)
+#define CIF_API __declspec( dllimport )
+#else
+#define CIF_API
+#endif
+#endif
+
 /* ----------------------------------------------------------------------
    CIFTI XML structure:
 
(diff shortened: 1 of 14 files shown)


seanm:
Nice!  You could also do similar for GCC/Clang: https://gcc.gnu.org/wiki/Visibility

Julianiolo:
@seanm yeah, I will look into that :)

seanm:
I'm just a nifti user, so it's just my 2¢.

I don't have Windows, but I can try building your patch on Mac...

seanm:
It doesn't compile because `ZNZ_API` is never defined. Likewise the others I suspect.

Easier to see if you indent the whole if/else chain that defines them.

Julianiolo:
@seanm Wow how did I not see that lol
Anyways, I now added the gcc/clang visibility attributes :)

cifti/afni_xml_io.h
```diff
@@ -50,26 +50,50 @@
...
 
+#ifndef CIF_API
+   #if defined(_WIN32) || defined(__CYGWIN__)
+      #if defined(CIFTI_BUILD_SHARED)
+      #ifdef __GNUC__
+         #define CIF_API __attribute__ ((dllexport))
+      #else
+         #define CIF_API __declspec( dllexport )
+      #endif
+      #elif defined(CIFTI_USE_SHARED)
```
seanm, line 61: I'm not sure these BUILD_SHARED / USE_SHARED conditions are needed... at least, other projects like [ITK](https://github.com/InsightSoftwareConsortium/ITK/blob/3ada77dd641c14e1dfd5a9a6f18f738eea148838/Modules/Core/Common/include/itkMacro.h#L210) and [VTK](https://github.com/Kitware/VTK/blob/e19328f6fce9a51f992fff5dbde5a3b8a82209ff/Common/Core/vtkABI.h#L37) don't seem to have them.

Seems would be simpler without them.

nifti2/nifti_tool.h
```diff
@@ -230,83 +254,83 @@ typedef struct {
...
+NI2_API int fill_ana_field_array(field_s * ah_fields);
+NI2_API int modify_all_fields(void *basep, nt_opts *opts, field_s *fields, int flen);
+NI2_API int modify_field     (void * basep, field_s * field, const char * data);
+NI2_API int process_opts     (int argc, char * argv[], nt_opts * opts);
+NI2_API int remove_ext_list  (nifti_image * nim, const char ** elist, int len);
+NI2_API int usage            (char * prog, int level);
+NI2_API int use_full         (void);
+NI2_API int verify_opts      (nt_opts * opts, char * prog);
+NI2_API int write_hdr_to_file (nifti_1_header * nhdr, const char * fname);
+NI2_API int write_hdr2_to_file(nifti_2_header * nhdr, const char * fname);
```
afni-rickr, line 324: These bits might not be needed for nifti_tool, as it isn't a library.

cifti/afni_xml_io.h
Julianiolo: Huh, it does seem to work using only dllexport

nifti2/nifti_tool.h
Julianiolo: That is true probably

cifti/afni_xml_io.h
Julianiolo: I'm not sure how this works exactly, e.g. [zlib](https://github.com/madler/zlib/blob/0f51fb4933fc9ce18199cb2554dacea8033e7fd3/zconf.h#L334) also seems to do it like I do (kinda, I think they define the ZLIB_INTERNAL inside their .c files when including the zlib.h header)

nifti2/nifti_tool.h
Julianiolo: @afni-rickr So do you mean the whole nifti_tool.h is not needed?

afni-rickr: @Julianiolo, I just mean that nifti_tool.h is only used in nifti_tool.c, which is a main program file.  It is not used as an external library, so I would not expect there to be a need for any dll or shared library controls.

cifti/afni_xml_io.h
seanm: Basically, it's harmless to attach the 'export' attribute in the case of building for static libraries, so simpler to always do it.

seanm:
I've just tried, and this builds successfully on macOS. I tried various permutations too, like debug, release, x86_64, arm64, toggling build options in cmake, etc.

nifti2/nifti_tool.h
Julianiolo: Done :)

hjmjohnson (approved):

